PR #218, opened 2016-08-12
Apply impulse force torque

pierredavidbelanger:
Some methods I am using in my project


pierredavidbelanger:
oups, wrong destination branch :) sorry

PR #219, opened 2016-08-12
add apply LinearImpulse / Force / ForceToCenter / AngularImpulse / Torque methods to PhysicsComponent

pierredavidbelanger:
Some methods I am using in my project

### DIFF
--- a/src/main/java/com/almasb/fxgl/physics/PhysicsComponent.java
+++ b/src/main/java/com/almasb/fxgl/physics/PhysicsComponent.java
@@ -28,6 +28,7 @@ package com.almasb.fxgl.physics;
 
 import com.almasb.ents.AbstractComponent;
 import com.almasb.ents.component.Required;
+import com.almasb.fxgl.app.FXGL;
 import com.almasb.fxgl.entity.component.BoundingBoxComponent;
 import com.almasb.fxgl.entity.component.PositionComponent;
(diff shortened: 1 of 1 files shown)


pierredavidbelanger:
I am not sur all the physics math are correct, since the forces are usually based on Newton metre and I am using [toMeters](https://github.com/AlmasB/FXGL/blob/master/src/main/java/com/almasb/fxgl/physics/PhysicsWorld.java#L580) to convert pixel to N-m ... but it works :)

AlmasB:
Yeah, technically it should, since there's a 1:1 correlation. Do you have any screenshots I could use for [showcase](https://github.com/AlmasB/FXGL/wiki/Sample-Projects)?

pierredavidbelanger:
Its only a toy project where agents evolve in a physic 2d world.
Their decisions (movement, attack, breed) are made using neural network.
When they breed, they produce new agents with a new neural network  evolved using the [NEAT](http://nn.cs.utexas.edu/?neat) algorithm.

I am more concern about the NeuralNet part than the visual part, so for now it looks like boring colored squares running everywhere on a boring white background :)

So, no screenshot so far, I will let you know when (if) I have some decent one to show.

Thank you!

AlmasB:
Sounds awesome! Let me know what kind of support you need from FXGL. Though it seems jbox2d is the most important aspect in your case.

PR #270, opened 2017-01-21
Issue 251

stefanbanu:


AlmasB:
It would be beneficial to have more concise calls, for example:
```java
if (PlatformUtils.isWindows()) {

} else if (PlatformUtils.isMac()) {

} else if (PlatformUtils.isLinux()) {

}
```

AlmasB:
looking good! just waiting on travis to build it

PR #426, opened 2017-11-12
Issue#386 (HitBox - generate name automatically if not provided)

SNException:
Hey again :)

The name of a HitBox is now optional and will be auto generated (hashCode of the given BoundingShape) if using the new overloaded constructor.


AlmasB:
Could you do the same for `HitBox(String name, Point2D localOrigin, BoundingShape shape)`

PR #501, opened 2018-02-25
Null issues resolved

Hanzallah:
Null Issues resolved


AlmasB:
looks good. thanks

Hanzallah:
Thanks. Can you tell me how to run your program because I have created jar files using the mvn clean package command but they are not running. Is there some step I am missing?

PR #561, opened 2018-06-29
reification

tieskedh:
Adds reifications at easy to implement places to provide more kotlin-like syntax


AlmasB:
I think this is has very little gain compared to the size of additions. Plus, for Java users it's invisible anyway.

I do like reification for Logger though, since most Logger instances are obtained in Kotlin (implementation side). If you want to keep all Logger reified calls and restore the rest, I'll be happy to merge this.

tieskedh:
what is the size of addition?
It's all inline what means zero overhead

AlmasB:
Since the API is used both from Java and Kotlin, I would prefer to keep both relatively similar until I am happy with the overall design of the engine and the way forward. Hence, I do not want to introduce anything that is Kotlin-specific, knowing that what exists can easily be removed from the engine.

My comment about Logger still stands since Logger is mostly going to be used from Kotlin.

PR #562, opened 2018-06-29
refactoring sequences

tieskedh:


AlmasB:
I believe some of these calls are not equivalent, can you confirm?

```
- for (i in entities.indices) {
-              val e = entities[i]
-              result.add(e)
- }

+            entities.forEach { result.add(it) }
```

In the first case the generated bytecode does not have any object allocations. In the 2nd case, `forEach()` is compiled to 

```
for (Iterator localIterator = $receiver$iv.iterator(); localIterator.hasNext(); result.add(it))
{
      Object element$iv = localIterator.next();it = (Entity)element$iv;
}
```

which for `ArrayList` implementation does create a new object, namely `new Itr()`, thus violating function doc that says "GC-friendly".

tieskedh:
unfortunately, your right...
It is only optimized with arrays.

PR #604, opened 2019-03-15
 Correction to the Contribution guide and  Adding the correct test for the getLengthAndNormalize function

BraddockG:


AlmasB:
Great! Thanks!

Issue #657, opened 2019-08-24
LiftComponent refactor


AlmasB:
Extract code below into `LiftData` private class to refactor 

```
    /**
     * Move in X axis.
     */
    var isHorizontal = false

    /**
     * Move in Y axis.
     */
    var isVertical = false

    var isGoingUp = false
    var isGoingRight = true

    var speedX = 0.0
    var speedY = 0.0

    var distanceX = 0.0
    var distanceY = 0.0

    var durationX: Duration = Duration.ZERO
    var durationY: Duration = Duration.ZERO

    private lateinit var timerX: LocalTimer
    private lateinit var timerY: LocalTimer
```

SNException:
@AlmasB 
Hello :)
Do you mind If I try to resolve this issue?

AlmasB:
Sure, go for it! Let me know if it's clear what needs to be done because I
don't quite remember what the issue is

PR #670, opened 2019-10-04
Issue #657 LiftComponent refactor

SNException:
Hi!
Let me know if this is how you intended it to be refactored.

I wrapped all related liftdata fields into one a "LiftData" class.

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/LiftComponent.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/LiftComponent.kt
@@ -6,6 +6,7 @@
 
 package com.almasb.fxgl.dsl.components
 
+import com.almasb.fxgl.core.math.Vec2
 import com.almasb.fxgl.dsl.FXGL
 import com.almasb.fxgl.entity.component.Component
 import com.almasb.fxgl.time.LocalTimer
@@ -20,13 +21,21 @@ class LiftComponent : Component() {
 
     private class LiftData() {
 
-        var isHorizontal = false
+        var isOn = false
         var isGoingPositive = true
         var speed = 0.0
         var distance = 0.0
         var duration: Duration = Duration.ZERO
 
         internal lateinit var timer: LocalTimer
+
+        fun update(tpf: Double): Double {
+            if (timer.elapsed(duration)) {
+                isGoingPositive = !isGoingPositive
+                timer.capture()
+            }
+            return if (isGoingPositive) speed * tpf else -speed * tpf
+        }
     }
 
     private val liftDataX = LiftData()
@@ -43,23 +52,14 @@ class LiftComponent : Component() {
     }
 
     override fun onUpdate(tpf: Double) {
-        if (liftDataX.isHorizontal) {
-            if (liftDataX.timer.elapsed(liftDataX.duration)) {
-                liftDataX.isGoingPositive = !liftDataX.isGoingPositive
-                liftDataX.timer.capture()
-            }
-
-            entity.translateX(if (liftDataX.isGoingPositive) liftDataX.speed * tpf else -liftDataX.speed * tpf)
+        val vector = Vec2()
+        if (liftDataX.isOn) {
+            vector.x = liftDataX.update(tpf).toFloat()
         }
-
-        if (!liftDataY.isHorizontal) {
-            if (liftDataY.timer.elapsed(liftDataY.duration)) {
-                liftDataY.isGoingPositive = !liftDataY.isGoingPositive
-                liftDataY.timer.capture()
-            }
-
-            entity.translateY(if (liftDataY.isGoingPositive) -liftDataY.speed * tpf else liftDataY.speed * tpf)
+        if (liftDataY.isOn) {
+            vector.y = liftDataY.update(tpf).toFloat()
         }
+        entity.translate(vector)
     }
 
     fun xAxisDistanceDuration(distance: Double, duration: Duration) = this.apply {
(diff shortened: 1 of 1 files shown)


AlmasB:
Hi, this is in the right direction. Can we apply stricter refactoring and get rid of repetition: currently the data for X and Y is repeated. Ideally we should have:

```
val liftDataX = LiftData()
val liftDataY = LiftData()

liftDataX.timer ..
liftDataX.isGoingPositive
liftDataX.duration

// etc.
```

I don't know if this can be refactored that way, but it should be possible given a lot of existing code duplication

SNException:
@AlmasB Thanks for the feedback!
I tried to make the LiftData object more general (see 2nd commit). Perhaps you can check again?

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/LiftComponent.kt
```diff
@@ -18,104 +18,95 @@ import javafx.util.Duration
...
-     * Move in X axis.
-     */
-    var isHorizontal = false
+    private class LiftData() {
 
-    /**
-     * Move in Y axis.
-     */
-    var isVertical = false
+        var isHorizontal = false
```
AlmasB: Yes, that's what I was thinking of. However, I think `isHorizontal` is not quite correct since we had both horizontal and vertical. So we could move in both horizontal and vertical directions. With the new code, we can only move in one or the other, is that true? Please double check.

We've replaced these with X and Y lift data. So I believe each lift data object should have whether it's on or not:

```
// on Update

if (liftDataX.isOn) {
            if (liftDataX.timer.elapsed(liftDataX.duration)) {
                liftDataX.isGoingPositive = !liftDataX.isGoingPositive
                liftDataX.timer.capture()
            }

            entity.translateX(if (liftDataX.isGoingPositive) liftDataX.speed * tpf else -liftDataX.speed * tpf)
        }

// then do the same with liftDataY ...
```

Even with above it looks like it can be improved since essentially the code is very similar. I wonder if there should be a method inside LiftData that does that ...

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/LiftComponent.kt
```diff
@@ -18,104 +18,95 @@ import javafx.util.Duration
...
     fun yAxisDistanceDuration(distance: Double, duration: Duration) = this.apply {
-        distanceY = distance
-        durationY = duration
-        speedY = distance / duration.toSeconds()
+        liftDataY.distance = distance
+        liftDataY.duration = duration
+        liftDataY.speed = distance / duration.toSeconds()
 
-        isVertical = true
+        liftDataY.isHorizontal = false
```
AlmasB: Please see earlier comment. This, I  think, should be `liftDataY.isOn = true`

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/LiftComponent.kt
SNException: I compared the current behavior and the one from the state before this PR with the BBoxSample app. And it looks the same.
The vertical movement was now done when `isHorizontal `was set to false. That is why I set it to `false` in the yAxis functions (for liftDataY object). But I agree that this is very confusing. Therefore I would rename this field like you suggested to `isOn` and then each instance has it's on state of it.
I hope I understood you correctly on this.

About having the `LiftData` object solve the code duplication in the update method of LiftComponent we could do something like this:

```java
// inside LiftData class...

        fun translateX(tpf: Double, entity: Entity) {
            if (isOn) {
                negateDirection()
                entity.translateX(if (isGoingPositive) speed * tpf else -speed * tpf)
            }
        }

        fun translateY(tpf: Double, entity: Entity) {
            if (isOn) {
                negateDirection()
                entity.translateY(if (isGoingPositive) speed * tpf else -speed * tpf)
            }
        }

        private fun negateDirection() {
            if (timer.elapsed(duration)) {
                isGoingPositive = !isGoingPositive
                timer.capture()
            }
        }
    }

```

Then we only have to call these functions in the update method like so: 
```java
// inside LiftComponent class...

    override fun onUpdate(tpf: Double) {
        liftDataX.translateX(tpf, entity)
        liftDataY.translateY(tpf, entity)
    }
```
I am not to sure that it is a good idea to let LiftData access the entity object now, though.
I am happy to hear your opinion on this.

AlmasB: > I hope I understood you correctly on this.

Yeah, I think so.

> I am not to sure that it is a good idea to let LiftData access the entity object now, though

I agree. How about this:

```
on update() {
    val v = Vec2()
    if (liftDataX.isOn) {
        v.x = liftDataX.update()
    }

    if (liftDataY.isOn) {
        v.y = liftDataY.update()
    }

    entity.transate(v)
}
```

and

```
LiftData::update() {
           if (timer.elapsed(duration)) {
                isGoingPositive = !isGoingPositive
                timer.capture()
            }

     return if (isGoingPositive) speed * tpf else -speed * tpf
}
```

You'd have to double check the logic (sorry, I really should have written a unit test for this), but architecture wise I think we achieved quite a nice result

SNException:
@AlmasB I like your suggestion. Perhaps you can check once again (see 3rd commit)? I checked behaviour again with the help of the BBoxSample and it looks like before the PR.

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/LiftComponent.kt
SNException: I agree.

Issue #673, opened 2019-10-05
NPE when entity factory implementing class is declared private


AlmasB:
1. Find the code that throws NPE when an entity factory class is declared private
2. Either fix NPE (so private factories are allowed) or throw a meaningful exception stating that private factories are not allowed. The former is preferred.

SNException:
Hi.
I tried to reproduce this error. However instead of a NullPointerException I am actually getting a IllegalAccessException.
```
java.lang.IllegalAccessException: class com.almasb.fxgl.core.reflect.ReflectionUtils (in module com.almasb.fxgl.core) cannot access a member of class tutorial.BreakoutApp$BreakoutFactory (in module samples.main) with modifiers "public"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:361)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:591)
	at java.base/java.lang.reflect.Method.invoke(Method.java:558)
	at com.almasb.fxgl.core/com.almasb.fxgl.core.reflect.ReflectionUtils.lambda$findMethodsMapToFunctions$1(ReflectionUtils.java:87)
	at com.sun.proxy.$Proxy3.apply(Unknown Source)
	at com.almasb.fxgl.entity/com.almasb.fxgl.entity.GameWorld$create$2.invoke(GameWorld.kt:398)
	at com.almasb.fxgl.entity/com.almasb.fxgl.entity.GameWorld$create$2.invoke(GameWorld.kt:33)
	at com.almasb.fxgl.core/com.almasb.fxgl.core.util.DSLKt.tryCatchRoot(DSL.kt:33)
	at com.almasb.fxgl.entity/com.almasb.fxgl.entity.GameWorld.create(GameWorld.kt:398)
	at com.almasb.fxgl.entity/com.almasb.fxgl.entity.GameWorld.spawn(GameWorld.kt:371)
	at com.almasb.fxgl.entity/com.almasb.fxgl.entity.GameWorld.spawn(GameWorld.kt:359)
	at com.almasb.fxgl.all/com.almasb.fxgl.dsl.FXGL$Companion.spawn(FXGL.kt:309)
	at com.almasb.fxgl.all/com.almasb.fxgl.dsl.FXGL.spawn(FXGL.kt)
	at samples.main/tutorial.BreakoutApp.initGame(BreakoutApp.java:82)
[...]

PR #674, opened 2019-10-06
Fix for Issue #673

SNException:
Hi.
As discussed in the Issue #673 here is the PR which solves that issue.
Tests are still passing.


AlmasB:
Great!

Issue #669, opened 2019-09-28
Add setWidthFromRatio and setHeightFromRatio to Settings


AlmasB:
Example usage: (when width or height is known and the other aspect needs to be set using a specific ratio)

```
settings.setWidth(800);
settings.setHeightFromRatio(16 / 9);
```

shubhankarb180:
Can I know more about the issue ?

AlmasB:
Sure, `Settings.kt` needs to be updated to have two new functions named above. The idea is that the width (or height) is set by user and they want to set the other aspect (height or width) using some ratio, such as 16 / 9, so the function:

```
setHeightFromRatio( ratio ) {
    height = ... calculated based on width
}

same for setWidthFromRatio ...
``

shubhankarb180:
I don't know much about kotlin, but I know how to create a function. So could you guide me where exactly I need to place the function?

AlmasB:
[Here](https://github.com/AlmasB/FXGL/blob/dev/fxgl/src/main/kotlin/com/almasb/fxgl/app/Settings.kt#L238), just above the `toReadOnly` func

nandhae:
I'd like to pick this if it's still up for grabs.

AlmasB:
I think maybe @shubhankarb180 is already working it? If not, feel free to add this.

nandhae:
@AlmasB Sure, Thanks. @shubhankarb180 Let me know here if you are not working on it and want me to pick this up.

shubhankarb180:
@nandhae You can work too. I was not able to add as I don't know much kotlin.

nandhae:
@AlmasB Raised a PR with two simple functions which takes ratio as a fraction. The clarifications i would be needing are,
1. Assumed that ratio would be fraction and did not try to read as string. Eg: "16:9". Please verify if this is okay.
2. Rounded result height and width to int. Worried if it will have precision errors.
3. Could not add unit tests, since I could not access these methods in "GameSettingsTest"

Any suggestions or directions on these points would help. Would be happy to refactor and enhance.

PR #675, opened 2019-10-07
Add set width and height from ratio methods #669

nandhae:
**What this PR does**

Adds two new methods to **Settings.kt** as per the request in issue #669 

**Dev Notes**

1. Local build successful

2. No Unit tests for new code

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/Settings.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/Settings.kt
@@ -32,6 +32,7 @@ import javafx.scene.paint.Color
 import javafx.stage.StageStyle
 import java.util.*
 import java.util.Collections.unmodifiableList
+import kotlin.math.roundToInt
 
 enum class MenuItem {
 
@@ -234,6 +235,14 @@ class GameSettings(
 
         var achievements: List<Achievement> = arrayListOf()
 ) {
+
+    fun setHeightFromRatio(ratio: Double) {
+        height = (width / ratio).roundToInt()
+    }
+
+    fun setWidthFromRatio(ratio: Double) {
+        width = (height * ratio).roundToInt()
+    }
 
     fun toReadOnly(): ReadOnlyGameSettings {
         return ReadOnlyGameSettings(
--- a/fxgl/src/test/kotlin/com/almasb/fxgl/app/GameSettingsTest.kt
+++ b/fxgl/src/test/kotlin/com/almasb/fxgl/app/GameSettingsTest.kt
@@ -102,4 +102,20 @@ class GameSettingsTest {
         assertFalse(settings.isLinux)
         assertFalse(settings.isMac)
     }
+
+    @Test
+    fun `Test setting of height and width from ratio`() {
+        val settingsInitial = GameSettings()
+
+        assertThat(settingsInitial.width, `is`(800))
+        assertThat(settingsInitial.height, `is`(600))
+
+        settingsInitial.width = 100;
+        settingsInitial.setHeightFromRatio(0.5);
+        assertThat(settingsInitial.height,  `is`(200))
+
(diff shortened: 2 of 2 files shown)


AlmasB:
Please raise the PR against `dev` branch, not `master`.

fxgl/src/main/kotlin/com/almasb/fxgl/app/Settings.kt
```diff
@@ -235,6 +236,14 @@ class GameSettings(
         var achievements: List<Achievement> = arrayListOf()
 ) {
 
+    public fun setHeightFromRatio(ratio: Double) {
```
AlmasB: 1. `public` is redundant, it's `public` by default. 
2. looks good, but unit tests are needed. You can access these new methods from [this line](https://github.com/AlmasB/FXGL/blob/dev/fxgl/src/test/kotlin/com/almasb/fxgl/app/GameSettingsTest.kt#L60). So first set width to a value, then call setHeightFromRatio and check the result. Then do the same with height. Finally, restore width and height to 500 so the other tests pass

nandhae: Changed the base to dev.
Forgot to remove the public keyword, it was a mistake.
I'll add the unit tests and commit.

AlmasB: Actually, scratch that last comment. It would be better to create one new test function:

```
fun `Test set ...`() {
 val settings = GameSettings()
// then do the test
}
```

nandhae: And i don't want to pollute beforeAll for testing this. I'll write as a separate test and initiate settings once again there if that's okay.

nandhae: Ignore my last reply too, explained the same without noticing your last reply.

nandhae:
@AlmasB Made appropriate changes and added tests. Had to force update the last commit to make it easy to review :)

fxgl/src/test/kotlin/com/almasb/fxgl/app/GameSettingsTest.kt
```diff
@@ -102,4 +102,15 @@ class GameSettingsTest {
         assertFalse(settings.isLinux)
         assertFalse(settings.isMac)
     }
-}
\ No newline at end of file
+
+    @Test
+    fun `Test setting of height and width from ratio`() { val app = MockGameApplication()
```
AlmasB: please delete `val app = MockGameApplication()`

fxgl/src/test/kotlin/com/almasb/fxgl/app/GameSettingsTest.kt
```diff
@@ -102,4 +102,15 @@ class GameSettingsTest {
...
+        val settingsInitial = GameSettings()
+
+        settingsInitial.setHeightFromRatio(0.5);
```
AlmasB: It would probably be clearer to set width explicitly before this line

fxgl/src/test/kotlin/com/almasb/fxgl/app/GameSettingsTest.kt
```diff
@@ -102,4 +102,15 @@ class GameSettingsTest {
...
+        assertThat(settingsInitial.height,  `is`(1600))
+
+        settingsInitial.setWidthFromRatio(0.5);
```
AlmasB: Perhaps a different ratio, so that the width is now different to default one?

nandhae:
@AlmasB Changed the test to check initial values and then tested the methods with different values.

AlmasB:
Great, thanks!

PR #682, opened 2019-11-11
Introduced a more generic bar view component

marvinbuff:
Not related to any open issue I implemented a new and more generic bar view component as a simple feature addition.


fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/view/GenericBarViewComponent.kt
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A general bar view component to display properties like health, mana, or energy.
+ * Can be fully customized using its [bar] and [maxValue] property.
+ * @author Marvin Buff (marvinbuff@hotmail.com)
+ */
+open class GenericBarViewComponent(x: Double, y: Double, valueProperty: SimpleIntegerProperty, color: Color) : ChildViewComponent(x, y, false) {
+
+    val bar = ProgressBar()
+    val maxValue = SimpleIntegerProperty(valueProperty.get())
```
AlmasB: I think we should expose `bar`'s current value and max value to avoid duplicate properties. So, maybe:
```
fun valueProperty(): DoubleProperty = bar.valueProperty()
fun maxValueProperty(): DoubleProperty = bar.maxValueProperty()
```

Then, the following is not needed:
```
        bar.currentValueProperty().bind(
                valueProperty.divide(maxValue).multiply(100)
        )
```

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/view/GenericBarViewComponent.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.almasb.fxgl.ui.ProgressBar
+import javafx.beans.property.SimpleIntegerProperty
+import javafx.scene.paint.Color
+
+/**
+ * A general bar view component to display properties like health, mana, or energy.
+ * Can be fully customized using its [bar] and [maxValue] property.
+ * @author Marvin Buff (marvinbuff@hotmail.com)
+ */
+open class GenericBarViewComponent(x: Double, y: Double, valueProperty: SimpleIntegerProperty, color: Color) : ChildViewComponent(x, y, false) {
```
AlmasB: instead of valuePropery, please add two `Double`s: currentValue and maxValue to ctor as primitives, not properties. Use these two values to set the bar's current and max values in `init {}`.

Please see the comment below wrt exposing bar's properties directly.

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/view/HealthBarViewComponent.kt
```diff
@@ -6,34 +6,16 @@
...
 import javafx.scene.paint.Color
 
 /**
  *
  * @author Almas Baimagambetov (almaslvl@gmail.com)
  */
-class HealthBarViewComponent(x: Double, y: Double) : ChildViewComponent(x, y, false) {
-
-    val hpBar = ProgressBar()
+class HealthBarViewComponent(x: Double, y: Double) : GenericBarViewComponent(x, y, SimpleIntegerProperty(2999), Color.GREEN) {
```
AlmasB: I think we should just delete this file. Users can easily create something similar with `GenericBarViewComponent` directly

marvinbuff:
I implemented the requested changes and added an utility constructor for when you already have a property which should just be displayed.

AlmasB:
A great PR, many thanks!

PR #685, opened 2019-11-21
Fix ProgressBar issue

marvinbuff:
I noticed a small issue when using the second constructor of the `GenericBarViewComponent`. 
It happened when you would initiate the class and bind its `currentValue` and then called the `setWidth` method. As far as I understood the code of `setWidth`, the width is set and then you marginally changed the currentValue to trigger its listener and therefore a repaint. Problem was that the unidirectionally bound property throws an error if changed manually, and I presume it would be worse in a bidirectional binding, as the actual value would be changed. 

These changes should hopefully fix it, so that I can conveniently use the class :)


marvinbuff:
I generally prefer to write in Kotlin so I try to convert files to Kotlin when working on them. Should I try to keep files in Java in FXGL?
But in particular, I did want to use the default values in the constructor of the `GenericBarViewComponent` without adding @JVMOverloads just for this example. 
Thinking about that, if you want better support for Java I should probably add the overload anyway.

AlmasB:
Yes jvmoverloads is a good idea for Java users.

Kt is fine, just place the file in src main kotlin under an equivalent
package

Many thanks

AlmasB:
Great, thanks!

PR #700, opened 2019-12-10
Added PropertyMapView along with a PropertyMapViewSample, closes #617

### LINKED ISSUE #617: Add PropertyMapView in the fxgl-ui module
An editable view for the `PropertyMap` data structure. This is a prerequisite for #615 

The view should look like a table or a grid with 2 columns: 1st column - property name, 2nd column - value. The values should be editable, which would directly propagate to the underlying PropertyMap.

CharlyZhu:


AlmasB:
Great PR, thanks!

Issue #368, opened 2017-08-04
Help Wanted: Unit Tests


AlmasB:
There is a lot of code that would benefit from having unit tests, non-UI stuff in particular

coldwarrl:
You wanna prioritize a little / give hints (which classes/functionality) in particular ? I guess I can help write some here and there...

AlmasB:
[codecov](https://codecov.io/gh/AlmasB/FXGL/tree/0.4.1/fxgl/src/main) should give you a good idea about what is covered and what isn't. I'd suggest we push those classes / functionality that are near 100% coverage to full 100% because we are somewhat confident these things are used. For those closer to 0% coverage I wouldn't bother for the time being.

Simulant87:
The link you posted is outdated. Here is the entry point into the dashboard overview [https://codecov.io/gh/AlmasB/FXGL](https://codecov.io/gh/AlmasB/FXGL)

AlmasB:
@Simulant87 Many thanks

AlmasB:
Most of the core modules is now tested. Unit tests for specific packages / files will be opened as separate issues as needed.

PR #782, opened 2020-04-06
Issue #368 unit tests - Array

ross-holloway94:
I wasn't quite sure how far to take this, but I have added a couple of unit tests for the Array class as per AlmasB's suggestion via email. Unit tests are a bit strange to do retrospectively in my opinion.

It would also be nice if a style guide / xml would be available somewhere so I wouldn't need to filter out the slight formatting changes that IntelliJ wanted to make (particularly spacing between some parameters and brackets).

Relates to issue #368


AlmasB:
Looks good. Many thanks!

AlmasB:
It seems non-public test methods are not accessible: https://travis-ci.org/github/AlmasB/FXGL/jobs/671703490

so I made them public. Odd that the PR didn't trigger CI...

ross-holloway94:
Huh that's strange. IntelliJ was able to run them just fine :/

Issue #785, opened 2020-04-10
ProgressBar customization


WojciechSerafin:
As we talk on Gitter that will be great if we can override ProgressBar with custom version to make it fit better to the game. If that’s not possible can you at least let developer access functions of ProgressBar’s rectangles options like corner arcs?

ross-holloway94:
@AlmasB Perhaps we could make `backgroundBar` and `innerBar` final to prevent them from being declared as a new object, but allow access to them?

AlmasB:
Adding two public accessor methods `getBackgroundBar()` and `getInnerBar()` seems like sufficient, at least in the meantime?

WojciechSerafin:
I’m sorry for late respond. Thats great news. It’s perfect solution for me
right now.

W dniu sob., 11.04.2020 o 14:17 Almas Baimagambetov <
notifications@github.com> napisał(a):

AlmasB:
@WojciechSerafin these methods are now available. 

To get the latest build (after CI has published them, which should be done in a few mins), you can use this:
```
<dependencies>
    <dependency>
        <groupId>com.github.almasb</groupId>
        <artifactId>fxgl</artifactId>
        <version>dev-SNAPSHOT</version>
    </dependency>
</dependencies>

<repositories>
    <repository>
        <id>oss.sonatype.org-snapshot</id>
        <url>http://oss.sonatype.org/content/repositories/snapshots</url>
    </repository>
</repositories>
```

ross-holloway94:
Is this issue now ready to be closed?

AlmasB:
I believe so.

PR #786, opened 2020-04-11
Issue #785 progressbar customization

ross-holloway94:
Allow access to backgroundBar and innerBar Rectangle objects so that they may be customised.

Signed-off-by: ross-holloway94 <ross.holloway94+github@gmail.com>


AlmasB:
Great, thanks Ross

Issue #772, opened 2020-03-15
Add virtual joystick input


AlmasB:
Visually, it should look similar to [this](https://github.com/MarcoFazioRandom/Virtual-Joystick-Godot#godot-virtual-joystick).

First, the API design should be proposed based on what information we want to obtain from the joystick. Presumably, we want a vector from the joystick center to where the player _pointer_ (finger / cursor) is, when the joystick is being pressed. Using the vector we can obtain both the direction and length (magnitude that we can use to estimate _force_).

Edit: functionality is there apart from force, which needs to be calculated and obtained via `vector` before this is closed.

AlmasB:
Initial API design complete, see [initial commit](https://github.com/AlmasB/FXGL/commit/c3ab02159bf3f39aa0198c0a1cbac0a6c3af8385) 

The commit includes two `TODO` placeholders, which primarily require UI design and implementation. See link in the description for visual reference (does not have to look exactly like that).

AlmasB:
@ross-holloway94 interested? There is also a sample included in the above commit to see how the virtual joystick would work.

ross-holloway94:
Yeah, I'll have a look this week and see if I can get my head around it

ross-holloway94:
Okay, initial thoughts are that the `VirtualJoystick` should instead be  composed of two `Circle` objects: A background and a foreground.
Then:

- The center of the foreground circle shall not be dragged past the edge of the background circle. If those points are equal, the maximum possible magnitude is returned.
- The foreground circle shall listen for dragging events
- On release, the foreground circle shall reset its position back to the center of the background circle.

AlmasB:
Yes, sounds good, you should be able to use the `pressedProperty()` of the `Parent` (which `VirtualJoystick` is) to see if the foreground circle should be moved to where player input (cursor or finger) is.

ross-holloway94:
Good stuff! :)

PR #787, opened 2020-04-21
#772 - Add Virtual Joystick input.

ross-holloway94:
WIP draft PR for virtual joystick.


AlmasB:
Thanks, please see some points below:

- Fine if you want to move to separate files, but I would keep the initial API, e.g. `fun createVirtualJoystick(): VirtualJoystick = FXGLVirtualJoystick(this)` for consistency. We can provide a different function in the future to allow users to configure their own `VirtualJoystick`

- `abstract class` to `interface` change: are you sure there is no common functionality we could keep in the abstract class? For example, I would have thought the following would be the same across any implementation of a `VirtualJoystick`.

```
val vector: Point2D
        get() = input.mousePositionUI.subtract(center.add(translateX, translateY))
```
- Good idea to expose inner and outer circles, though I would keep them as `Node`, since users may provide a lot more complex views than just `Circle`.

```
val outerCircle: Circle
    val innerCircle: Circle
```

- Not sure about the public API `fun addInnerCircleListener()`, shouldn't the listener be always present / active from the moment the joystick was created? Perhaps another point in favour of having `abstract class`, so we can hide impl details, e.g. `protected fun`.

- ` fun reset()` sounds good. Is this to place the inner circle into default position when the user lifts their finger?

Looking forward to seeing the next version.

AlmasB:
Looks good. To track the vector from center point to user finger, you can use the previous code or something similar. I suspect you will need to have access to `Input` to get the information about where mouse cursor (user finger) is.

```
abstract class VirtualJoystick(private val input: Input) : Parent() {

    /**
     * A point in local coordinates that defines the center of this joystick.
     * It is used for computing the vector.
     */
    abstract val center: Point2D

    val vector: Point2D
        get() = input.mousePositionUI.subtract(center.add(translateX, translateY))
}
```

PR #803, opened 2020-05-07
add Entity.getComponent() extension method which uses type inference

wakingrufus:
add Entity.getComponent() extension method which uses type inference instead of a class parameter


AlmasB:
Many thanks

Issue #788, opened 2020-04-27
Design and implement an inventory API


AlmasB:
First, the design should be proposed.

Some thoughts to include: inventory item quantities (always 1, like weapons/armour in Van Helsing or possible stack with quantities like in Resident Evil 3). Both should be supported.

A sort button.

Views: regardless of the underlying inventory API, we should provide a default (customizable) UI view.

AlmasB:
Extracted view issue into #957 . The API is complete.

PR #831, opened 2020-06-11
Initial attempt to the Inventory functionality. #788

CharlyZhu:


AlmasB:
Thanks! The tentative view API is now in `sandbox/inventory`.

PR #832, opened 2020-06-15
add functions to AssetLoaderService to check if an asset exists

wakingrufus:
add functions to AssetLoaderService to check if an asset exists in order to support dynamic logic when loading assets

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/services/AssetLoaderService.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/services/AssetLoaderService.kt
@@ -119,6 +119,18 @@ class AssetLoaderService : EngineService() {
         }
     }
 
+
+
+    fun assetExists(assetName: String): Boolean{
+        return getAssetFromCache(assetName) != null
+                || FXGLApplication.app.javaClass.getResource(assetName)
+                // try /fxglassets/ from fxgl.all module using this javaclass
+                ?: javaClass.getResource("/fxgl${assetName.substring(1)}") != null
+    }
+    fun textureExists(name: String): Boolean = assetExists(TEXTURES_DIR + name)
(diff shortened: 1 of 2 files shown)


AlmasB:
Could you provide more details on the use case?

Current design of the asset loader is based on the fact that the requested asset always exists at runtime. If it does not, then that's a developer error. Therefore, there is no need to check if an asset exists.

wakingrufus:
Sorry, I left a note in the gitter, but I should have also put more details here.
My use case is that I want to load animations made from _n_ number of discrete images, without having know what _n_ is up front in every case.
Here is my current sample code: https://github.com/wakingrufus/praxis/blob/master/game/src/main/kotlin/com/github/wakingrufus/rpg/AssetUtil.kt
Right now, i have to attempt to load each image (letting an exception occur), then check if the result is the dummy image. 
this seems like it would be cleaner (and not generate the extra WARN logs) with minimal changes to FXGL code, but I am sure a more elegant solution could be designed if you want this to be a first class supported use case.

AlmasB:
Thanks for the details. This is usually handled with some metadata available upfront for each animation, which conforms to the design above (i.e. developer knows what assets are available). I appreciate that it may not always be beneficial to generate or store such metadata, so I suggest that you check for asset (I guess "texture" specifically) availability directly in your `AssetUtil`, something along these lines:

```
javaClass.getResource("/assets/textures/$assetName") != null
```

Based on the above design, I am not ready to merge this yet.

wakingrufus:
I think i will switch to LPC spritesheets instead, since that is somewhat of a standard, and the framework supports it better:
https://github.com/jrconway3/Universal-LPC-spritesheet

AlmasB:
Yeah it's a nice sprite sheet generator. I'm using it for my rpg game too: https://www.youtube.com/watch?v=L6A6n1kqDCE

PR #834, opened 2020-06-19
Make TMX level parser more robust by removing whitespace.

dykstrom:
I propose to make the TMX level parser more robust by trimming strings before converting them to integers in parseData. Someone might do what I did, edit the TMX file by hand and let IntelliJ format it...


AlmasB:
Great patch! Many thanks

PR #840, opened 2020-07-07
completed RechargeableDoubleComponentTest, resolves #833

### LINKED ISSUE #833: Complete RechargeableComponent and its test
Code coverage: https://codecov.io/gh/AlmasB/FXGL/src/dev/fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/RechargeableComponents.kt

File in fxgl module: RechargeableDoubleComponentTest

dykstrom:
Please have a look at test cases "Negative damage" and "Change max" that fail if you uncomment the asserts. There are different ways to solve these problems, but I guess you have to decide how you want it to work. You could for example disallow negative damage, but then there might be a cursed weapon that actually does negative damage... You could of course also do some more min/max calculations to ensure value is never > maxValue.

### DIFF
--- a/fxgl/src/test/kotlin/com/almasb/fxgl/dsl/components/RechargeableDoubleComponentTest.kt
+++ b/fxgl/src/test/kotlin/com/almasb/fxgl/dsl/components/RechargeableDoubleComponentTest.kt
@@ -6,8 +6,10 @@
 
 package com.almasb.fxgl.dsl.components
 
+import javafx.beans.value.ChangeListener
 import org.hamcrest.CoreMatchers.`is`
 import org.hamcrest.MatcherAssert.assertThat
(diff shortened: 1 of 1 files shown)


AlmasB:
Many thanks for this. Good point about value / max value.

1. I think dealing negative damage is application level logic. So my inclination is towards letting the user figure out what they want to do and then either damage() or restore() based on their game logic. On that basis, dealing and restoring negative values shouldn't make sense, so I will probably make these functions treat negative values as 0.
2. Semantically, it makes a lot of sense for `value` to be <= `max value` at all times, so I will embed this relation into the component.

Let me know what you think.

AlmasB:
Consistency is indeed key. In these cases, FXGL's policy is to inform the user but keep going until we can't. I'll make the necessary changes. Thanks for the feedback.

PR #851, opened 2020-07-30
#520 DSL: for each var type add a new Text object generation bound to the var

### LINKED ISSUE #520: DSL: for each var type add a new Text object generation bound to the var

AahzBrut:
Add function addPropText to FXGL, which creates Text with text value of the PropertyMap element.
And also done a little housekeeping:

single object imports in files which I edited
suppress Kottlin warnings about dependencies in test classes
edit deprecated annotation to make SonarLint happy
change deprecated getGameState() to getWorldProperties()
closes #520


fxgl/src/main/kotlin/com/almasb/fxgl/dsl/FXGL.kt
```diff
@@ -589,6 +606,39 @@ class FXGL private constructor() { companion object {
...
+                    .also { addUINode(it, x, y) }
+    }
+
+    /**
+     * Returns property value as String
+     * @param propertyName name of the property in PropertyMap
+     * @return string value of property
+     * @throws IllegalArgumentException if property has unknown type (unknown in terms of 'when' clauses)
+     */
+    @JvmStatic private fun getPropertyValueAsString(propertyName: String): String? {
```
AlmasB, line 631: This should return `StringBinding` or `StringProperty`, i.e. the new text object's `textProperty()` needs to be bound to a property variable, so that when the variable value changes, the text is updated too.

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/FXGL.kt
```diff
@@ -589,6 +606,39 @@ class FXGL private constructor() { companion object {
...
 
+    /**
+     * Returns new Text initialized with property value, whose name was passed to the function
+     * @param propertyName name of the property in PropertyMap
+     * @param x x coordinate
+     * @param y y coordinate
+     * @throws IllegalArgumentException if property was not found in PropertyMap
+     * @return Text()
+     */
+    @JvmStatic fun addPropText(propertyName: String, x: Double, y: Double) : Text {
```
AlmasB, line 617: This should be renamed to `@JvmStatic fun addVarText(varName: String, x: Double, y: Double): Text`. There is already a function with this signature. That function (which only worked for integer properties) should be replaced with this new one.

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/FXGL.kt
AlmasB: I think the fix is to just remove the `.get()` calls?

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/FXGL.kt
AahzBrut: ok, done

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/FXGL.kt
AahzBrut: changed return type to StringExpression for  `newText(StringExpression textBinding)`

PR #853, opened 2020-07-30
#699 Add PropertyMap::forEach

### LINKED ISSUE #699: Add PropertyMap::forEach
PropertyMap is a data structure located in the fxgl-core module. The forEach function should act in the same way as the JDK's forEach of a Map: https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#forEach-java.util.function.BiConsumer-

Essentially, it should iterate over all key-value pairs in the map and run a bi-consumer on them.

This addition would also require adding a test to PropertyMapTest to ensure it is fully covered.

The call site (from the user's perspective):

```
map.forEach((k, v) -> ... );
```

AahzBrut:
add forEach function to PropertyMap which iterate through all elements and call BiConsumer on each of them

add test for forEach function to check its working as intended 

closes #699

### DIFF
--- a/fxgl-core/src/test/kotlin/com/almasb/fxgl/core/collection/PropertyMapTest.kt
+++ b/fxgl-core/src/test/kotlin/com/almasb/fxgl/core/collection/PropertyMapTest.kt
@@ -253,23 +253,23 @@ class PropertyMapTest {
         assertTrue(s.contains("testInt=IntegerProperty [value: 3]"))
     }
 
-    private var timesCounter = 0
-    private var sumCounter = 0
-
     @Test
     fun `forEach test`() {
+        var timesCounter = 0
+        var sumCounter = 0
+
         map.setValue("testElement1", 2)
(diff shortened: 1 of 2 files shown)


fxgl-core/src/test/kotlin/com/almasb/fxgl/core/collection/PropertyMapTest.kt
```diff
@@ -251,6 +253,25 @@ class PropertyMapTest {
         assertTrue(s.contains("testInt=IntegerProperty [value: 3]"))
     }
 
+    private var timesCounter = 0
```
AlmasB, line 256: Please move `timesCounter`, `sumCounter` and `counter(key: String, value: Any)` to the scope of `forEach test` to be consistent with existing tests where similar approach was used.

AahzBrut: fixed

PR #862, opened 2020-08-01
#790 Add acceleration to ProjectileComponent

### LINKED ISSUE #790: Add acceleration to ProjectileComponent
The new addition should be accompanied by an associated test

AahzBrut:
Add acceleration to projectile component

Add test to ensure that acceleration is working as intended

closes #790


fxgl/src/test/kotlin/com/almasb/fxgl/dsl/components/ProjectileComponentTest.kt
```diff
@@ -34,6 +35,40 @@ class ProjectileComponentTest {
...
+        val acceleration = Point2D(100.0, 0.0)
+
+        comp = ProjectileComponent(Point2D(0.0, 0.0), 0.0, acceleration)
+        e.addComponent(comp)
+
+        assertThat(comp.acceleration, `is`(acceleration))
+
+        comp.onUpdate(tpf)
+
+        assertThat(comp.direction, `is`(acceleration.normalize()))
```
AlmasB, line 63: 1. Please update assertions of `direction` and `speed`, using hardcoded expected values to be consistent with other tests, e.g.:

```
assertThat(comp.direction, `is`(Point2D(1.0, 0.0)))
```

Same for lines 69-70.

2. We also need to test the entity position, after each `comp.onUpdate()`, e.g.:

```
assertThat(e.position, `is`(Point2D(1.0, 0.0)))
```

AahzBrut: Done

AahzBrut:
there was an error in position calculation, fixed

PR #865, opened 2020-08-02
#824 Update built-in components with isComponentInjectionRequired()

### LINKED ISSUE #824: Update built-in components with isComponentInjectionRequired()
We now have `boolean isComponentInjectionRequired()` that can be overridden by subclasses of Component. There are many built-in components that FXGL provides. Ideally all of them should be tweaked so no injection is required and returning false for the above method to speed up component addition.

See [this commit](https://github.com/AlmasB/FXGL/commit/b4909b5ab119f2d569aa97968ef1d7a8a9f8f689) for some examples. Some components already do not require injection can override the above method immediately.

AahzBrut:
Add initialization of child components in onAdded method; 
Add implementation of isComponentInjectionRequired() with return value of false to all descendants of Component class;

closes #824


fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/components/DataComponents.kt
```diff
@@ -39,6 +39,8 @@ abstract class BooleanComponent
     }
 
     override fun toString() = "${javaClass.simpleName.substringBefore("Component")}($value)"
+
+    override fun isComponentInjectionRequired(): Boolean = false
```
AlmasB, line 43: I think we should not add this to any abstract Component in this file as we don't know whether the subclasses will need injection.

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/view/GenericBarViewComponent.kt
```diff
@@ -52,4 +52,5 @@ open class GenericBarViewComponent @JvmOverloads constructor(
     fun valueProperty(): DoubleProperty = bar.currentValueProperty()
     fun maxValueProperty(): DoubleProperty = bar.maxValueProperty()
 
+    override fun isComponentInjectionRequired(): Boolean = false
```
AlmasB, line 55: Same comment here, this is an open class, so a subclass may still require injection

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/components/DataComponents.kt
AahzBrut: Fixed

fxgl/src/main/kotlin/com/almasb/fxgl/dsl/components/view/GenericBarViewComponent.kt
AahzBrut: Fixed

PR #867, opened 2020-08-02
#660 InjectInTest::inject use names provided by the @Inject() annotation to inject values

### LINKED ISSUE #660: InjectInTest::inject use names provided by the @Inject() annotation to inject values
Current the call site is:

```
inject(lookup, achievementManager, "eventBus",  EventBus())
inject ...
```

What it should be:

```
val map = mapOf<String, Any>( "eventBus" to EventBus(), ...  );
inject(lookup, achievementManager, map)
```

Effectively reducing number of calls to `inject` to 1. The `inject` method should use the `@Inject` provided name to find the correct value from the map.

AahzBrut:
Add convenience inject method to InjectInTest which receives Map of parameters to inject;

Replace multiple injects in tests with new method;

closes #660


AlmasB:
Thanks!

PR #880, opened 2020-09-29
Clean up code in AStarPathfinder

### LINKED ISSUE #794: Clean up code in AStarPathfinder
- [ ] No need for a separate `findPath` method at Line 51. Should be merged back to the L33 or L38 methods.
- [ ] `getValidNeighbors` at L135 should make use of already existing method `grid.getNeighbours` and then filter as needed.

marvinbuff:
Resolves #794

I did not yet merge the `findPath` methods as for me it makes sense to have a pure function with the actual implementation. Should I do it regardless?

`getValidNeighbours` should be fine like this.


AlmasB:
Thanks, any thoughts why tests pass on jdk14 and ea, but not 11: [travis ci](https://travis-ci.org/github/AlmasB/FXGL/builds/731325137)?

marvinbuff:
I am not sure. I think it might not be deterministic. The test now fails locally for me as well, which it did at some point but not when I pushed. What is especially weird is that the test fails which does not have busy cells. 
I will investigate.

Would it bother you if I convert some files to Kotlin?

marvinbuff:
I think I understand the issue, my change made the solution path non-deterministic. If you look at the example below with the start state s, goal state g, and two other states x and y.
s | x 
y | g 
Both paths `s-x-g `and `s-y-g` are valid, it just depends which state is expanded first. This non-deterministic character was introduced by either the stream or the set. 
I think set is not appropriate here as we generally care for order in search. The stream limitation I would like to remove by just going to Kotlin.

AlmasB:
Nice detective work!

> This non-deterministic character was introduced by either the stream or the set.

This got me curious, so I looked up their [documentation](https://docs.oracle.com/javase/8/docs/api/?java/util/stream/package-summary.html).

> Streams may or may not have a defined encounter order. Whether or not a stream has an encounter order depends on the source and the intermediate operations. Certain stream sources (such as List or arrays) are intrinsically ordered, whereas others (such as HashSet) are not.

The source (grid.getNeighbors()) is a List, so `stream()` will keep the order, but the implementation of `toSet()` is a `HashSet`, which, according to its doc, says:

> It makes no guarantees as to the iteration order of the set;

fxgl-ai/src/test/java/com/almasb/fxgl/pathfinding/astar/AStarPathfinderTest.java
```diff
@@ -80,8 +79,9 @@ public void testFindPathWithBusyCells() {
                 3, 4,
                 4, 4,
                 4, 5);
-        path = pathfinder.findPath(1, 1, 4, 5, Collections.singletonList(grid.get(3, 4)));
-        assertPathEquals(path,
+        List<AStarCell> pathWithBusyCell = pathfinder.findPath(1, 1, 4, 5, Collections.singletonList(grid.get(3, 4)));
+        System.out.println("Solution: " + pathWithBusyCell);
```
AlmasB, line 83: Could we revert changes to the test file. I don't think you meant to commit these.

AlmasB (changes requested):

fxgl-ai/src/test/java/com/almasb/fxgl/pathfinding/astar/AStarPathfinderTest.java
marvinbuff: I actually meant to, because I prefer to have meaningful output in the test so you actually now what the error is. Because if the error says: 2 not equal 1, then I will not know anything. 

But of course I can revert this if it doesn't fit your approach.

marvinbuff:
> The source (grid.getNeighbors()) is a List, so `stream()` will keep the order, but the implementation of `toSet()` is a `HashSet`, which, according to its doc, says:

Streams actually keep order. Good to know. Thanks.

fxgl-ai/src/test/java/com/almasb/fxgl/pathfinding/astar/AStarPathfinderTest.java
AlmasB: Good point about output. Adding `System.out` here would be inconsistent compared to other tests. So I think the test can be improved, particularly the bit that compares paths. Can we rewrite that, so that `assertThat` correctly prints the output when failing? Something like `assertThat(list1, is(list2))`?

marvinbuff:
I improved the reporting of the error and converted the tests to Kotlin. Is this alright?

fxgl-ai/src/test/java/com/almasb/fxgl/pathfinding/astar/AStarPathfinderTest.kt
```diff
@@ -0,0 +1,96 @@
...
+                2, 5,
+                2, 6,
+                3, 6,
+                4, 6,
+                4, 5)
+    }
+
+    private fun assertPathEquals(path: List<AStarCell>, vararg points: Int) {
+        val pointsList = points.toList().chunked(2) { it[0] to it[1] }
+        assert(pointsList.size == path.size) { reportNotMatchingPaths(path, pointsList) }
```
AlmasB, line 83: According to [assert doc](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/assert.html) these are only active when `-ea` JVM option is enabled.

If the above is true, please replace all `assert` with `assertEquals`, which can also store a failure message.

AlmasB (changes requested):

fxgl-ai/src/test/java/com/almasb/fxgl/pathfinding/astar/AStarPathfinderTest.kt
marvinbuff: Is it alright to use the Kotlin `assertEquals`? Not sure I added the dependency correctly.

marvinbuff: image
You don't know by chance why I get this warning? I guess my setup is somehow wrong.
Anyway, running works fine so there is no issue.

AlmasB: It's an IDE bug, I've also got it, just ignore it.

AlmasB:
Good stuff, many thanks!

PR #882, opened 2020-10-11
Refactor NineSliceTextureBuilder

### LINKED ISSUE #799: Refactor code in NineSliceTextureBuilder::build()
There is a lot of (repetitive?) code in build(), which can potentially be cleaned up.

marvinbuff:
Resolves #799

Small refactoring of the `NineSliceTextureBuilder`. I slightly changed the style of all methods to what I seem is a more readable alternative. Obviously that is a matter of taste and I can revert it if you prefer the alternative.

Please consider adding the `hacktoberfest-accepted` tag if the contribution was helpful.


AlmasB:
Thanks for this. I won't have time to review this until the weekend, but if you want to speed this up...

We'll need a test eventually for this builder. So what I'm thinking is:

1. Create a WxH image using current implementation of the builder and save as .png (using UINineSliceSample in fxgl-samples)
2. Write a test that uses same WxH to produce a new image and compare pixel by pixel to the one produced in (1)
3. Commit this (can use a different PR if that's helpful for your hacktoberfest event)
4. Test this new implementation

If you are not interested, I'm happy to do the above over the weekend.

marvinbuff:
Sounds interesting, I will give it a shot. Unfortunately, I will probably not have time over the week as well. So no hurry from my side.

I think the best would be to create a second issue and handle it independently from this here. Then I will try to resolve it over the weekend and you can just review this PR on its own.

AlmasB:
Have you tried running the #886 test against this new implementation?

marvinbuff:
> Have you tried running the #886 test against this new implementation?

Yes. Runs successfully.

fxgl-media/src/main/kotlin/com/almasb/fxgl/texture/NineSliceTextureBuilder.kt
```diff
@@ -30,202 +31,115 @@ class NineSliceTextureBuilder(val texture: Texture) {
...
+            val rem = height.toInt() % base.height.toInt()
 
             for (i in 1 until times) {
-                newRight = newRight.superTexture(right, VerticalDirection.DOWN)
+                adjusted = base.superTexture(base, VerticalDirection.DOWN)
             }
 
             if (rem > 0) {
-                newRight = newRight.superTexture(right.subTexture(Rectangle2D(0.0, 0.0, right.width, rem.toDouble())), VerticalDirection.DOWN)
+                adjusted = base.superTexture(base.subTexture(Rectangle2D(0.0, 0.0, base.width, rem.toDouble())), VerticalDirection.DOWN)
```
AlmasB, line 120: should this be `adjusted = adjusted.superTexture ... `?

marvinbuff: Yes, thanks. Fixed!

AlmasB:
Thanks

PR #887, opened 2020-10-21
Add ability to stop all sounds and music

JWill18:
Adds some methods to pause/stop/resume all sounds and music that is currently playing. Useful for when you need to pause and turn off all noises and then need to resume any active sounds when unpausing.

### DIFF
--- a/fxgl-media/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
+++ b/fxgl-media/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
@@ -83,6 +83,13 @@ class AudioPlayer : EngineService() {
     }
 
     /**
+     * Stops all sounds from playing
+     */
+    fun stopAllSounds() {
+        activeSounds.forEach { stopSound(it) }
+    }
+
+    /**
      * Plays given music based on its properties.
      *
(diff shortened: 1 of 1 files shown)


fxgl-media/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
```diff
@@ -138,4 +145,35 @@ class AudioPlayer : EngineService() {
...
+     * Can be restarted by [.resumeMusic] and [.resumeAllMusic]
+     */
+    fun pauseAllMusic() {
+        activeMusic.forEach { pauseMusic(it) }
+    }
+
+    /**
+     * Start playing all active music if its not currently playing.
+     */
+    fun playAllMusic() {
```
AlmasB, line 160: Very nice, thanks.

Based on the doc of `pauseAllMusic()`, I think  `resumeAllMusic`, which would call `resumeMusic`, is missing.

JWill18: Oops, I accidently left that method out. Fixed it.

AlmasB:
Looks good, thanks

PR #888, opened 2020-10-23
Add pause callbacks to EngineService

### LINKED ISSUE #879: Add onMainLoopPausing onMainLoopResumed callbacks to services
1. `EngineService` needs to be updated with two new `open fun` named as above.
2. These two functions need documentation. Particularly we need to note _when_ the callbacks are called: the pausing is called before pause, but resumed is after resume.
3. In `Engine.kt`:

```
    fun pauseLoop() {
        // here

        loop.pause()
    }

    fun resumeLoop() {
        loop.resume()
        // here    
    }
```

need to call services with these new functions as appropriate as shown above.

marvinbuff:
Resolves #879

Should resolve the mentioned issue. I tested it manually and it worked. Of course those pause functions are only called when the window is minimized.

Not sure there is anything to test. It would make sense to me to add tests in the implementing class once those functions are used. Otherwise let me know.

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/core/EngineService.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/core/EngineService.kt
@@ -53,12 +53,12 @@ abstract class EngineService : Updatable, SerializableType {
     open fun onExit() { }
 
     /**
-     * Called just before the main loop is paused, which is only happens when the main window is minimized.
+     * Called just before the main loop is paused.
      */
     open fun onMainLoopPausing() { }
 
     /**
-     * Called just after the main loop is resumed, which is only happens when the minimized main window is opened.
+     * Called just after the main loop is resumed.
      */
     open fun onMainLoopResumed() { }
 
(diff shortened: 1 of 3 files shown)


fxgl-core/src/main/kotlin/com/almasb/fxgl/core/EngineService.kt
```diff
@@ -52,6 +52,16 @@ abstract class EngineService : Updatable, SerializableType {
      */
     open fun onExit() { }
 
+    /**
+     * Called just before the main loop is paused, which is only happens when the main window is minimized.
```
AlmasB, line 56: > which is only happens when the main window is minimized.

I think we should remove this bit, because in the future it may not be true. Plus, it refers to main window being minimized, but on mobile it would be called when the app enters a paused state. So just leave:

`Called just before the main loop is paused.`

fxgl-core/src/main/kotlin/com/almasb/fxgl/core/EngineService.kt
```diff
@@ -52,6 +52,16 @@ abstract class EngineService : Updatable, SerializableType {
...
+     */
+    open fun onMainLoopPausing() { }
+
+    /**
+     * Called just after the main loop is resumed, which is only happens when the minimized main window is opened.
```
AlmasB, line 61: Same comment as above.

AlmasB (changes requested):
Looks good, please have a look at the 2 comments

marvinbuff:
Thanks, fixed both comments. 

Would be nice if you could add the `hacktoberfest-accepted` label if you're happy with the changes.

AlmasB:
Great stuff, thanks!

PR #911, opened 2020-12-06
Subscene demo code added.

SergeMerzliakov:
A basic demo showing how to create subscenes and navigate between them.

### DIFF
--- a/fxgl-samples/src/main/kotlin/sandbox/subscene/SubSceneSampleApp.kt
+++ b/fxgl-samples/src/main/kotlin/sandbox/subscene/SubSceneSampleApp.kt
@@ -1,0 +1,64 @@
+/*
+ * FXGL - JavaFX Game Library. The MIT License (MIT).
+ * Copyright (c) AlmasB (almaslvl@gmail.com).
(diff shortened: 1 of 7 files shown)


fxgl-samples/src/main/kotlin/sandbox/subscene/SubSceneSampleApp.kt
```diff
@@ -0,0 +1,64 @@
...
+
+		val handler: EventHandler<NavigateEvent> = EventHandler { e: NavigateEvent? ->
+			handleNavigate(e!!)
+		}
+
+		FXGL.getEventBus().addEventHandler(NAVIGATION, handler)
+	}
+
+	override fun initUI() {
+		// TODO - is this the correct way to set up a SubScene on startup
```
AlmasB, line 41: At the moment, yes. I can't remember why -- I don't think it's a limitation, probably suboptimal handling of scenes during app initialization. Would be nice if we could push subscenes without having to wait for a new frame / tick.

AlmasB (approved):
Many thanks!

PR #970, opened 2021-02-20
Read tileset attribute margin from tmx file.

### LINKED ISSUE #969: TMXLevelLoader does not read tileset margin
**Describe the bug**
The TMXLevelLoader does not read the "margin" attribute of the "tileset" element in the tmx file. This leads to tiles looking strange if the tileset does in fact have a margin. The margin attribute should be read like the spacing attribute.

**To Reproduce**
1. Create a tmx file that uses a tileset image that has margins.
2. Load tmx file in FXGL application.
3. See that tiles look strange, for example 1 pixel off if the margin is 1.

**Expected behavior**
The margin attribute is read from the file and stored in Tileset.margin.

**Runtime info(please complete the following information):**
 - Windows 10, but the error is OS independent.

dykstrom:
Added code to read the margin attribute of the tileset element in tmx files. Resolves #969. The tileset image that has been added comes from https://opengameart.org/ and has a public domain license.


AlmasB:
Given there are no changes to TilesetLoader, I take it that tile map generation was correct, but the "margin" for tileset wasn't parsed correctly?

AlmasB (approved):

dykstrom:
Yes, as far as I know that is the case. I tried my example game with a locally built version of FXGL and the tiles looked correct. I have only tried with one example though, and that tileset image had a margin of 1.

PR #972, opened 2021-02-20
Created functionality for managing stacks in the inventory.

adambocco:
@AlmasB 

I made a bit of changes to Inventory.kt to allow setting of maximum stack size. This included adding additional data structures, classes, and methods to manage the stacks. Such as:

   - Creating StackData class for keeping track of individual stacks.
   - Modifying incrementQuantity to dynamically create stacks.
   - Create data structures to store and access these stacks in Inventory and ItemData classes.
   - Added public methods incrementStack and removeStack to directly manage stacks.
  
The private data structures I added in Inventory.kt allow storing individual stacks which themselves have an ID. This ID, which is used as the key to stacksData, may be useful in the future when creating an improved default view for the inventory which would allow moving/splitting stacks by tracking each stack's current 'slot' in the inventory.

Let me know what you think! 

Thanks,
Adam Bocco

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/Settings.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/Settings.kt
@@ -274,14 +274,15 @@ class GameSettings(
 
         var engineServices: MutableList<Class<out EngineService>> = arrayListOf(
                 // this is the order in which services will be initialized
-                // by design, the order of services should not matter
+                // by design, the order of services should not matter,
+                // however some services can depend on others, so no-dep ones should come first
+                FXGLAssetLoaderService::class.java,
                 FXGLApplication.GameApplicationService::class.java,
                 FXGLDialogService::class.java,
                 IOTaskExecutorService::class.java,
                 EventBusService::class.java,
                 FileSystemService::class.java,
                 LocalizationService::class.java,
-                FXGLAssetLoaderService::class.java,
                 SystemBundleService::class.java,
                 SaveLoadService::class.java,
                 FXGLUIFactoryServiceProvider::class.java,


fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
```diff
@@ -26,70 +27,148 @@ class Inventory<T>(var capacity: Int) {
...
-            view: Node = EmptyView,
-            quantity: Int = 1
+    @JvmOverloads
+    fun add(
+        item: T,
+        name: String = "",
+        description: String = "",
+        view: Node = EmptyView,
+        quantity: Int = 1,
+        stackMax: Int = 10
```
AlmasB: stack max should be `Int.MAX_VALUE` by default

fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
```diff
@@ -26,70 +27,148 @@ class Inventory<T>(var capacity: Int) {
...
+
     private val itemsProperty = FXCollections.unmodifiableObservableList(items)
 
+    private val stacks = FXCollections.observableArrayList<StackData<T>>()
+
+    private val stacksProperty = FXCollections.unmodifiableObservableList(stacks)
+
     private val itemsData = hashMapOf<T, ItemData<T>>()
 
+    private val stacksData = hashMapOf<Int, StackData<T>>()
```
AlmasB: I currently don't see why we we need an `Int -> Data` mapping. I thought we could simply add a `Map<T, List<ItemData<T> >>`. I haven't given this much thought, so I don't know if this works or not, but can you consider this option?

If the above works, then we don't really need `Stack` and any of its related code, right? Then 

`fun incrementQuantity(item: T, amount: Int): Boolean` is just going to:

```
if (item in map) {
    var list = map.get(item)
   // find first item data in list that can hold more qty or add new ItemData
} else {
    add new ItemData
}
```

I will review the rest once we have resolved this.

AlmasB (commented):
Thanks, please can we discuss the outstanding comments.

fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
adambocco: The `Int -> Data` mapping was unnecessary but I thought it might help when creating a grid-style view if each stack had an ID representing it's current slot in the inventory.  I created the StackData class to represent individual stacks and to not have repeat attributes from ItemData (each stack of same item will have the same description, stackMax, name, etc..). Additionally I was using the StackData type to select specific stacks in the InventoryView  with:

```private val stacks = FXCollections.observableArrayList<StackData<T>>()```

Either way I am starting from scratch with your comments in mind and I'm not sure if storing maxStack in ItemData is possible.  If `isRemoveItemIfQty0 == false` and maximum stack is an attribute on the `ItemData` this may result in `T -> empty ObservableList` with no reference to what the stackMax was. 
   - I can create a `map( T -> Int)` where Int is stackMax for each item, T.
   - I can create a make itemsData values int a pair to track stackMax:
       ```private val itemsData = hashMapOf<T, Pair<Int, ItemData<T>>>()``` 



Or I can simplify my original commit without IDs for each stack.

Let me know if I'm way off here, thanks again for the input.

AlmasB: I see, thanks for the explanation. 

I can see now how `Int -> Data` would be useful for that, but there is currently no use case for this. In the long term it would be good to have something like `StackData`. I appreciate the future-proof solution here, but incremental changes are easier to manage.  So I'd propose to tweak the current PR to minimise changes, whilst introducing this new feature.

> If isRemoveItemIfQty0 == false and maximum stack is an attribute on the ItemData this may result in T -> empty ObservableList with no reference to what the stackMax was.

Good point. Can we decompose the problem to one single (universal) max quantity variable. This will simplify the current PR. We can then discuss how we can handle max quantity on a per-item basis.

In the next (separate) PR we can consider how to clean up the implementation, including "repeat attributes from ItemData", for now `List<ItemData>` should be sufficient I believe.

PR #973, opened 2021-02-22
Created functionality for specifying inventory maximum stack size

adambocco:
@AlmasB 

I have simplified my PR to allow setting `stackMax` using the original design. In addition I have updated the inventory tests, sample, and default list view. In Inventory.kt, everything works the same if no `stackMax` is set. However, the user must index the item they want with `itemData[0]` to grab the first stack. 

In order to accomodate stackMax: 

  - I added a safeguard for setting stackMax if there is an item over that quantity already in inventory.
  - I added `getItemQuantity` to return the total quantity of all stacks of a specific item
  - `add()` and `incrementQuantity()` now dynamically create stacks within the bounds of `capacity`

Another thing to mention is that in order to create new stacks, I have to use some existing item as a template as seen in `incrementQuantity`. This means an item could be added of the same type `T` and have varying names and descriptions. 

Let me know what you think!

Thanks again,
Adam Bocco

### DIFF
--- a/fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
+++ b/fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
@@ -19,6 +19,7 @@ import javafx.scene.Node
  * Represents an inventory (bag) of user-defined items.
  *
  * @author Almas Baimagambetov (almaslvl@gmail.com)
+ * @author Adam Bocco (adam.bocco@gmail.com)
  */
 class Inventory<T>(var capacity: Int) {
 
--- a/fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
+++ b/fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
@@ -65,10 +65,12 @@ class Inventory<T>(var capacity: Int) {
     fun setStackMax(newStackMax: Int): Boolean {
         if (newStackMax >= stackMax) {
             stackMax = newStackMax
+            return true
         }
         for (item in allData.values) {
             for (stack in item) {
-                if (stack.quantity > newStackMax) return false
+                if (stack.quantity > newStackMax)
+                    return false
             }
         }
         stackMax = newStackMax
@@ -195,7 +197,6 @@ class Inventory<T>(var capacity: Int) {
         if (amount < 0 && itemQuantity + amount < 0)
             return false
 
-
         quantityCount = amount
 
         /**
@@ -215,7 +216,6 @@ class Inventory<T>(var capacity: Int) {
                     quantityCount += stack.quantity
 
                     stack.quantity = 0
-
                 }
 
                 if (isRemoveItemsIfQty0 && stack.quantity == 0) {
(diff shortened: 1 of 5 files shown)


fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
```diff
@@ -28,32 +29,65 @@ class Inventory<T>(var capacity: Int) {
...
+
+    private var stackMax = Int.MAX_VALUE
 
     /**
      * Number of items in inventory.
      */
     val size: Int
-        get() = items.size
+        get() {
+            var totalSize = 0
```
AlmasB, line 42: I'm unsure here. Should we return the number of items as before and have a separate `numStacks`, which would return number of stacks?

fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
```diff
@@ -64,41 +98,75 @@ class Inventory<T>(var capacity: Int) {
...
+
+        }
     }
 
     /**
      * Removes [item] from inventory, reducing the inventory size by 1.
      * If you only want to reduce quantity, use [incrementQuantity] with negative amount instead.
      */
-    fun remove(item: T) {
+    fun removeItem(item: T) {
```
AlmasB: Not sure if we need this API change here? Perhaps we could keep `remove()`.

fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
```diff
@@ -64,41 +98,75 @@ class Inventory<T>(var capacity: Int) {
...
         itemsData -= item
     }
 
-    fun getData(item: T): Optional<ItemData<T>> {
+    private fun removeStack(itemData: ItemData<T>) {
+        if (itemData.userItem in items)
+            itemsData[itemData.userItem]?.remove(itemData)
+    }
+
+    fun getData(item: T): Optional<ObservableList<ItemData<T>>> {
```
AlmasB: I think we can return an empty list here, to be consistent with the rest of the project.

AlmasB (changes requested):
Looks good, please see additional comments

fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt
adambocco: I could make that change but as the PR is, it would let the user add an arbitrary amount of stacks. `capacity` would be only a bound for number of item types, `T`, in the inventory. It would also make code like this in `incrementQuantity` unnecessary:

https://github.com/AlmasB/FXGL/blob/1c45b84d837322b9db72e4a68982ef544b7b06cd/fxgl-trade/src/main/kotlin/com/almasb/fxgl/inventory/Inventory.kt#L184-L194

AlmasB: I see, in which case we will need to clearly define what `size` and other parameters are in the context of Inventory. Consider this comment resolved.

adambocco:
@AlmasB 
I wasn't sure if I should append my name to the author section when making changes like this. If you would like me to, is there a certain documentation style I can follow to add that or an example file in FXGL with 2+ authors?

AlmasB:
```
@author name
@author name

etc.
```

Order doesn't matter. It makes sense to append author tag when making significant changes to the file.

AlmasB (approved):
Looks good, many thanks

Issue #966, opened 2021-02-13
Add isLoggingEnabled to EventBus


AlmasB:
Currently, EventBus will log each event that was fired. This can become problematic when the game has lots of events. Using this boolean flag, (set to true by default to be consistent with current implementation), developers can control this behaviour.

AlmasB:
closed by #979

PR #979, opened 2021-03-03
#966 Added isLoggingEnabled to EventBus

wolches:
Hello!

Not sure if my code are breaking any accepted practices or code conventions.
Is it ok?

### DIFF
--- a/fxgl-events/src/main/kotlin/com/almasb/fxgl/event/EventBus.kt
+++ b/fxgl-events/src/main/kotlin/com/almasb/fxgl/event/EventBus.kt
@@ -24,6 +24,8 @@ class EventBus {
 
     private val eventHandlers = Group()
 
+    var isLoggingEnabled = true
+
     /**
      * Register [eventHandler] for [eventType].
      */
@@ -48,7 +50,9 @@ class EventBus {
      * i.e. synchronous.
      */
     fun fireEvent(event: Event) {
-        log.debug("Firing event: $event")
+        if (isLoggingEnabled) {
+            log.debug("Firing event: $event")
+        }
 
         eventHandlers.fireEvent(event)
     }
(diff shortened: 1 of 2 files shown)


fxgl-events/src/main/kotlin/com/almasb/fxgl/event/EventBus.kt
```diff
@@ -24,6 +24,8 @@ class EventBus {
 
     private val eventHandlers = Group()
 
+    private var loggingEnabled = true
```
AlmasB: 1. Please remove `private`, Kotlin will auto-generate `getter/setter` pair, so the two new functions below are no longer needed.
2. Please rename variable name to `isLoggingEnabled` to preserve consistency in the codebase (all boolean variables begin with `is`).

fxgl-events/src/main/kotlin/com/almasb/fxgl/event/EventBus.kt
```diff
@@ -48,8 +50,18 @@ class EventBus {
...
     fun fireEvent(event: Event) {
-        log.debug("Firing event: $event")
+        if (loggingEnabled) {
+            log.debug("Firing event: $event")
+        }
 
         eventHandlers.fireEvent(event)
     }
+
+    fun isLoggingEnabled(): Boolean {
```
AlmasB: These two functions are not needed after the above change (removing `private`) is made.

AlmasB (changes requested):
Looks good, please consider the comments attached.

fxgl-events/src/main/kotlin/com/almasb/fxgl/event/EventBus.kt
wolches: Fixed

fxgl-events/src/main/kotlin/com/almasb/fxgl/event/EventBus.kt
wolches: Fixed

AlmasB (approved):
Great, many thanks!

PR #986, opened 2021-03-07
Created support for loading .tmx files with flipped tiles

adambocco:
@AlmasB 
I have done a few things in this PR:

  - Changed storage of .tmx data to Long to support flipped orientation encoding of .tmx data
  - Added CSV and base64 support for reading stream of UInt into Long.
  - Added support for flipping tiles using the new `flipHorizontally()` and `flipVertically()` functions. I made the check for diagonal flip as suggested but left the implementation blank.
  - Added a sample with examples of flipped and rotated tiles and objects in the sandbox.
  - Added a test for flipped tiles that checks the location of a black pixel depending on the tile's orientation.

Let me know how this looks.

Thanks!
Adam Bocco


fxgl-samples/src/main/java/sandbox/TiledFlippingSample.java
```diff
@@ -0,0 +1,61 @@
+package sandbox;
```
AlmasB, line 7: Missing license header

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TilesetLoader.kt
```diff
@@ -134,6 +145,20 @@ class TilesetLoader(private val map: TiledMap, private val mapURL: URL) {
 
             log.debug("Writing to buffer: dst=${x * map.tilewidth},${y * map.tileheight}, w=$w,h=$h, src=$srcx,$srcy")
 
+            if (isHorizontallyFlipped) {
+                sourceImage = flipHorizontally(Texture(sourceImage).subTexture(Rectangle2D(srcx.toDouble(), srcy.toDouble(), w.toDouble(), h.toDouble())).image)
```
AlmasB: I assume here `sourceImage` refers to the entire tileset image and not a single tile image. Is this why we need to do the Texture() subTexture() trick?

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TilesetLoader.kt
```diff
@@ -89,7 +92,15 @@ class TilesetLoader(private val map: TiledMap, private val mapURL: URL) {
 
         for (i in 0 until layer.data.size) {
 
-            var gid = layer.data.get(i)
+            val tempGid = layer.data.get(i)
+
+            // Orientation is encoded in 3 most significant bits of '.tmx' UInt which is embedded in lower half of Long
+            val isHorizontallyFlipped = tempGid and 0x80000000L != 0L
```
AlmasB: Could we use this format, for consistency?

https://github.com/AlmasB/FXGL/blob/6924aea33ec92a8f6eea9da3adfb637ba9208706/fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TMXLevelLoader.kt#L392

So 

```
val isHorizontallyFlipped = tempGid and (1L shl 31) != 0L
```

Then we can refer to tmx documentation and it's clearer which bit is being checked.

Same comment for checks below.

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TilesetLoader.kt
```diff
@@ -134,6 +145,20 @@ class TilesetLoader(private val map: TiledMap, private val mapURL: URL) {
...
+                srcx = 0
+                srcy = 0
+            }
+            if (isVerticallyFlipped) {
+                sourceImage = flipVertically(Texture(sourceImage).subTexture(Rectangle2D(srcx.toDouble(), srcy.toDouble(), w.toDouble(), h.toDouble())).image)
+                srcx = 0
+                srcy = 0
+            }
+            if (isDiagonallyFlipped) {
+                // Not currently supported
```
AlmasB: How about `log.warn(" ... ")` that it is not supported

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TMXLevelLoader.kt
```diff
@@ -334,40 +331,39 @@ class TMXLevelLoader : LevelLoader {
...
-                        val baos = ByteArrayOutputStream()
+                var inputStream: InputStream = InputStream.nullInputStream()
 
-                        InflaterInputStream(bytes.inputStream()).use {
-                            it.copyTo(baos)
-                        }
+                val longArray = arrayListOf<Long>()
 
-                        bytes = baos.toByteArray()
+                when (start.getString("compression")) {
```
AlmasB: How about:

```
val inputStream = when ( ... ) {
    // ...
    else -> {
         log.warn ( .. );
         InputStream.nullInputStream()
    }
}
```

Also, we should close the stream, I didn't catch if we are doing that ...., so:

`inputStream.use {  }` would be a safer option, or does this input stream come from memory, then it's fine I guess, please check.

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TMXLevelLoader.kt
```diff
@@ -334,40 +331,39 @@ class TMXLevelLoader : LevelLoader {
...
 
-                when (start.getString("compression")) {
-                    "zlib" -> {
-                        val baos = ByteArrayOutputStream()
+                var inputStream: InputStream = InputStream.nullInputStream()
 
-                        InflaterInputStream(bytes.inputStream()).use {
-                            it.copyTo(baos)
-                        }
+                val longArray = arrayListOf<Long>()
```
AlmasB: I think this can be moved after `when`.

AlmasB (changes requested):
Thanks, please see comments below.

fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TilesetLoader.kt
adambocco: Yes this is why, but only when the `Tileset.isSpriteSheet` is true. I can fix this to simply flip when loading from a collection of images and grab this subTexture and then flip when loaded from sprite sheet.

AlmasB (approved):
Looks great, many thanks

Issue #961, opened 2021-02-05
Warn and ignore (instead of crashing) if an entity is being to the game world more than once


AlmasB:
This is at GameWorld.kt::L73 in fxgl-entity. Also we should do something similar for removeEntity L108 same file.

Tests will probably need updating too.

oseph:
Perhaps in the EntityFactory documentation there could be mention that buildAndAttach should not be used in a call to `spawn()`? I think part of what confused me (as a n00b) is that in the current documentation I couldn't really tell the two methods apart.

AlmasB:
Thanks for the feedback, I will update the documentation as part of this
issue

AlmasB:
@adambocco could you take a look at this one please, when you've got time.

adambocco:
@AlmasB 
Definitely!

It's a busy week for me but after Saturday I will have much more time.

Thanks!

AlmasB:
Closed by #988

PR #988, opened 2021-03-14
Updated GameWorld.kt to warn instead of crashing on adding/removing entities

adambocco:
@AlmasB 

I updated `GameWorld.kt` to log warnings instead of crashing and updated the tests. Pretty simple but I think this is all you wanted for issue [961](https://github.com/AlmasB/FXGL/issues/961).

Let me know, thanks!
Adam Bocco


fxgl-entity/src/test/kotlin/com/almasb/fxgl/entity/GameWorldTest.kt
```diff
@@ -57,9 +57,12 @@ class GameWorldTest {
         val newWorld = GameWorld()
         newWorld.addEntity(e)
 
-        assertThrows(IllegalArgumentException::class.java, {
```
AlmasB, line 60: I think we should also rename the function, from `Throw` to `Ignore`, as with the other one.

AlmasB (changes requested):
Looks good, just one comment

AlmasB (approved):
Great, many thanks

PR #990, opened 2021-03-17
Changed Logger::warning to print full stack trace.

adambocco:
After looking over some of the examples of using `warning(message, error)` I think this change is good, definitely helps to have a stack trace over just the exception type. If you think it becomes too lengthy in situations where many warnings with errors are logged we can truncate the trace but I don't particularly want to do that.

Thanks!
Adam Bocco


AlmasB (approved):
Great, thanks!

AlmasB:
For future reference, please can you include `Closes # issue-number` in the PR comment box -- this should automatically close the associated issue

Issue #900, opened 2020-11-22
Improve Tiled map loading performance


AlmasB:
- [ ] ~~The tile generation could potentially be parallelized.~~ parallel tile loading didn't yield significant improvements in performance.
- [x] Remove debug logs that begin `log.debug("Writing to buffer: ....`

AlmasB:
@adambocco apart from above (parallelizing tile generation -- I still need to investigate this), any thoughts on how we can improve loading performance of .tmx?

adambocco:
@AlmasB 
I spent some time looking at this. It seems like when a call is made to `setLevelFromMap('mapName.tmx')` there can be a couple seconds of load time on larger maps. I did some tests with a 500x500 tiled map of 16x16 tile sizes (tile layer entities) and I found that there is quite a slowdown in `TilesetLoader.kt::loadView(layerName: String)` here:
https://github.com/AlmasB/FXGL/blob/74c9c8cf4609c510b366456a4fc9caec80067019/fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/level/tiled/TilesetLoader.kt#L158-L163

Commenting out the line calling `log.debug(..)` gave me a big increase in performance in this specific case:

- Parse Time: 0.1913669
- TilesetLoader Init Time: 0.0024212
- Load View (for loop): 0.2882948
- Pixel Writer Time (consider overhead of 'measureNanoTime'): 0.1251819
- Load Image Time (consider overhead of 'measureNanoTime'): 0.0387208
- Conditional Time (consider overhead of 'measureNanoTime'): 0.0663441
- Initial Time (consider overhead of 'measureNanoTime'): 0.0374719
- **Log Flip Time (consider overhead of 'measureNanoTime'): 0.0080994**
- LoadView Time: 0.483584
-    Create Tile Layer Entities Time: 0.4860122
-    Create Object Layer Entities Time: 0.0425438
- **<------ setLevelFromMap timed ------>**
- **Engine Time: 0.7000000000000005**


Compared to:


- Parse Time: 0.2222178
- TilesetLoader Init Time: 0.0642288
- Load View (for loop): 1.933842
- Pixel Writer Time (consider overhead of 'measureNanoTime'): 0.2197043
- Load Image Time (consider overhead of 'measureNanoTime'): 0.0701937
- Conditional Time (consider overhead of 'measureNanoTime'): 0.1095921
- Initial Time (consider overhead of 'measureNanoTime'): 0.0730058
[...]

PR #1013, opened 2021-04-19
#900 Added tiled benchmarking sample in sandbox

adambocco:
I added some of the maps I have been using to test performance and a sample class for testing them. 

I included some maps with entirely flipped tiles which has terrible performance (as you can imagine) over ones with unflipped tiles. This could probably be improved, but in actual use I don't know if flipped tiles are common enough to warrant it.

I would include the 'closes' tag but I wanted to take a last look at any way I could improve `loadView` and looking at alternative buffers.

Can I ask what you were using to run your benchmarks in your comment a couple days ago?

Thanks!
Adam Bocco


AlmasB:
The Tiled sample I ran is in `sandbox/benchmark`.

As for flipped tiles, yeah I think you are right, we should focus on the common use case first and only then worry about the other 10-20%.

fxgl-samples/src/main/java/sandbox/tiled/TiledBenchmarkSample.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    protected void initGame() {
+        getGameWorld().addEntityFactory(new TiledObjectFactory());
+
+        var t1engine = getEngineTimer().getNow();
+        var t1sys = System.nanoTime();
+
+        setLevelFromMap("tmx/benchmarking/csv_1layer_noflip_500x500.tmx");
```
AlmasB: Set level from map can give misleading times as it also includes the call to `gameworld.setLevel`, which in turn forces a scene graph update. I also made this mistake too at first :)

Can you replace `setLevelFromMap` to just load the level. You can check the Tiled sample from sandbox/benchmark.

Incidentally, since all benchmark code is in there, perhaps we can move this file there too. I'm thinking then `sandbox/tiled` is for developing Tiled related features. See what you think

AlmasB (changes requested):
Just a minor comment.

fxgl-samples/src/main/java/sandbox/tiled/TiledBenchmarkSample.java
adambocco: I made those fixes and created a simple UI to run the tests although I'm not sure if that was wise. Would that interfere much with the times for the tests themselves even if on separate threads? 

 I didn't notice the place for benchmarking samples earlier. I left the empty tiled package in sandbox as well, I can see myself adding something to there in the future.

AlmasB: Haven't checked yet, but if you start the benchmark tests using `getExecutor().startAsync()`, it should be fine. Essentially, any thread, apart from UI thread, is fine.

AlmasB:
@adambocco 

Given there is a UI and everything now, do you want to move my tmx there too and delete my sample?

Edit: with your setup of List<String>, you probably won't even need to move the tmx, just add a file name.

AlmasB (approved):

PR #1025, opened 2021-04-24
Added more syntax highlighting to readme

cegredev:
Added syntax highlighting for the Maven and Gradle snippets. It appears GitHub does not support Java modules syntax yet, but in case it ever will, I added the right language to that code block as well.


AlmasB (approved):
Looks good, thanks

PR #1028, opened 2021-04-27
Removed Swing dependency

FDelporte:
Proposal to remove Swing from the project as it was only used for image conversion

### DIFF
--- a/fxgl/src/main/java/module-info.java
+++ b/fxgl/src/main/java/module-info.java
@@ -21,10 +21,10 @@ module com.almasb.fxgl.all {
     requires transitive javafx.graphics;
     requires transitive javafx.controls;
     requires transitive javafx.fxml;
-    requires transitive javafx.swing;
 
     requires com.fasterxml.jackson.databind;
     requires com.gluonhq.attach.lifecycle;
+    requires java.desktop;
 
     opens com.almasb.fxgl.dsl to com.almasb.fxgl.core;
     opens com.almasb.fxgl.dsl.components to com.almasb.fxgl.core;
--- a/fxgl/src/main/java/module-info.java
+++ b/fxgl/src/main/java/module-info.java
@@ -24,7 +24,6 @@ module com.almasb.fxgl.all {
 
     requires com.fasterxml.jackson.databind;
     requires com.gluonhq.attach.lifecycle;
-    requires java.desktop;
 
     opens com.almasb.fxgl.dsl to com.almasb.fxgl.core;
     opens com.almasb.fxgl.dsl.components to com.almasb.fxgl.core;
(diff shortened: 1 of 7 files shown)


fxgl/src/main/kotlin/com/almasb/fxgl/app/FXGLApplication.kt
```diff
@@ -60,6 +59,12 @@ import java.nio.file.Paths
 import java.time.LocalDateTime
 import javax.imageio.ImageIO
 import kotlin.system.measureNanoTime
+import java.io.ByteArrayInputStream
```
AlmasB: Are these two new imports necessary? Would be great to not have anything `java.awt` in imports.

fxgl/src/main/kotlin/com/almasb/fxgl/app/FXGLApplication.kt
```diff
@@ -670,7 +675,7 @@ class FXGLApplication : Application() {
          */
         fun saveScreenshot(): Boolean {
             val fxImage = mainWindow.takeScreenshot()
-            val img = SwingFXUtils.fromFXImage(fxImage, null)
+            val img = ImageIO.read(ImageIO.createImageInputStream(fxImage))
```
AlmasB: Do we know what the difference is in the result between these two calls?

fxgl/src/main/java/module-info.java
```diff
@@ -21,10 +21,10 @@
     requires transitive javafx.graphics;
     requires transitive javafx.controls;
     requires transitive javafx.fxml;
-    requires transitive javafx.swing;
 
     requires com.fasterxml.jackson.databind;
     requires com.gluonhq.attach.lifecycle;
+    requires java.desktop;
```
AlmasB: Do we need an explicit dependency on java.desktop?

AlmasB (commented):
Thanks, this is a good idea to remove javafx.swing. Please can you resolve the comments below.

FDelporte:
To my surprise turning a JavaFX image into a file seems to be more complicated than I hoped... I made a test project on https://github.com/FDelporte/JavaFxTestImageConversion which uses a solution proposed on StackOverflow after trying out many other (unsuccessful) approaches... Not really sure if this is the correct way forward?!

AlmasB:
I would probably use the reference SwingFXUtils implementation from javafx.swing and try to adapt it. Does the SwingFXUtils conversion have any dependencies? If not, then it should reasonably easy to extract the necessary bits.

FDelporte:
License is OK to be copied/modified/included?
https://github.com/openjdk/jfx/blob/master/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java

AlmasB:
Unfortunately, no, GPL2 allows private use, but for releases, the entire project needs to be converted to GPL2 as far as I understand.

Edit: I just read the "classpath" exception bit, which _does_ seem to allow use of the code in a project with a different but compatible license. Anyway this makes no difference since we now have our own implementation.

There might be a different way to convert fx Image to BufferedImage. I wonder if we can get the underlying buffers, similar to how the particles are drawn in [FXGL-FastRender](https://github.com/AlmasB/FXGL-FastRender) then just re-copy pixels. I'll give this a go and will keep you updated.

AlmasB:
I merged your refactor code with FXGL-FastRender implementation of buffer transfer in [commit](https://github.com/AlmasB/FXGL/commit/737c2ba44b9c3e89ed29137ebe7f978b10ac782b)

Issue #993, opened 2021-03-27
Identify all places where exceptions are thrown within FXGL that can be replaced with a warning


AlmasB:
To fix, all files and lines where the above happens should be reported here. Ctrl + Shift + F in IntelliJ can be used to search for the "throw" keyword.

AlmasB:
Closed in 11.17

PR #1033, opened 2021-04-29
 ignore and warn adding duplicate component, #993

Zhack47:


AlmasB:
Well spotted, thanks.

@adambocco please can you review when you get a chance

fxgl-entity/src/test/kotlin/com/almasb/fxgl/entity/EntityTest.kt
```diff
@@ -49,7 +49,7 @@ class EntityTest {
 
     @Test
     fun `Add component fails if same type already exists`() {
-        val exception = assertThrows(IllegalArgumentException::class.java) {
+        val exception = assertDoesNotThrow {
```
adambocco, line 52: The test name doesn't match the test performed in this change. Should we just get rid of this test and make one like `Adding duplicate component is ignored` and check if the original component remains on the entity.

fxgl-entity/src/test/kotlin/com/almasb/fxgl/entity/EntityTest.kt
```diff
@@ -170,14 +170,15 @@ class EntityTest {
         assertThat(entity.components, hasItem(control))
     }
 
+    // Test seems duplicate, see l.51 `Add component fails if same type already exists`
```
adambocco: Same here. Maybe change this to create two `TestControl` objects and check if the first one is the added component like:
```
control1 = TestControl()
control2 = TestControl()
entity.addComponent(control1)
entity.addComponent(control2)
assertThat(entity.getComponent(TestControl::class.java),  `is`(control1))
```

adambocco (changes requested):
The change looks good as long as it seems like the best idea to warn and ignore instead of throwing an error for duplicate components. I could see a case for either side but would rather the ignore as well.

fizietechD (approved):

fizietechD (approved):

fxgl-entity/src/test/kotlin/com/almasb/fxgl/entity/EntityTest.kt
AlmasB: Good point, updated.

AlmasB (approved):
Thanks for the reviews @adambocco @Fizie92 

Thanks for the PR @Zhack47

PR #1047, opened 2021-05-21
Get size of EntityGroup

adambocco:
Seems like a simple change.
I added a getter for the size of the entities copy list, added a test for it, and cleaned up the old tests.


AlmasB (approved):
Looks great, thanks!

PR #1059, opened 2021-05-29
Updated Physics Collision Sample and added new Keybinding Closes #965

### LINKED ISSUE #965: Add player grow-shrink controls to PhysicsCollisionSample
The sample is in fxgl-samples /   sandbox: https://github.com/AlmasB/FXGL/blob/dev/fxgl-samples/src/main/java/sandbox/PhysicsCollisionSample.java

We need to add a few `onKeyDown()` that will enlarge / shrink the player by changing its scale. So we can check various collision situations. The ability to rotate the player would also be useful here.

jo372:


AlmasB (approved):
Many thanks

PR #1060, opened 2021-05-29
ignore and warn updating component

lydianeU:


AlmasB (approved):
Looks good, many thanks.

PR #1061, opened 2021-05-29
added setters for hearing sense component properties

jo372:


AlmasB (approved):
Looks good, many thanks

Issue #1100, opened 2021-09-08
EntityDrawingOrderSample is wrong - doesnt use zindex


mortensen:
The current example EntityDrawingOrderSample doesnt use the setZIndex() methods. 

**To Reproduce**
Run the code, press F. Wont switch the z index of the squares.

**Expected behavior**
The red and black squares should switch places.

I will send a pull request to fix the example.

PR #1101, opened 2021-09-08
https://github.com/AlmasB/FXGL/issues/1100 - Issue 1100. Fixing zindex

mortensen:
Fixed the example. Uses zindex now.


AlmasB (approved):
Looks good, many thanks

PR #1179, opened 2022-06-04
Enable loading of @2x images for high-dpi displays

rolve:
@AlmasB I struggled a lot to create games with high-res images using manual scaling. Then I found this: https://dlsc.com/2017/08/29/javafx-tip-27-hires-retina-icons/

The important part is: "_When you want to load images in code you have to make sure not to use input streams but a URL_"


AlmasB:
Many thanks for this.

PR #1195, opened 2022-08-07
build: Kotlin 1.6.0 -> 1.7.10, JavaFX 18 -> 18.0.2

Anivie:
Kotlin has preview the Kotlin K2 compiler,it can help us to reduce build time,but fxgl don't support kotlin 1.7+,so I think we need to update it.

Update JavaFx Version to 18.0.2
Update Kotlin Version to 1.7.10
Fix:Kotlin Unsupported Kotlin 1.7+

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
@@ -78,8 +78,10 @@ internal class ObservableTrigger(trigger: Trigger) {
 }
 
 data class KeyTrigger
-@JvmOverloads constructor(val key: KeyCode,
-                          override val modifier: InputModifier = InputModifier.NONE) : Trigger {
+@JvmOverloads constructor(
+    val key: KeyCode,
+    override val modifier: InputModifier = InputModifier.NONE
+) : Trigger {
 
     override val name: String = key.getName()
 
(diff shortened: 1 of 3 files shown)


fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
```diff
@@ -78,8 +78,10 @@ internal class ObservableTrigger(trigger: Trigger) {
 }
 
 data class KeyTrigger
-@JvmOverloads constructor(val key: KeyCode,
-                          override val modifier: InputModifier = InputModifier.NONE) : Trigger {
```
AlmasB, line 82: I am not sure we need to make any changes to this file?

fxgl/src/main/kotlin/com/almasb/fxgl/dev/Console.kt
```diff
@@ -112,7 +112,9 @@ class Console : Pane() {
     }
 
     private fun cmd(name: String, description: String, function: (Array<String>) -> Unit, syntax: String = "") {
-        commands.put(name, Command(name, description, Consumer { function.invoke(it) }, syntax))
```
AlmasB, line 115: Same comment as above, not sure why we need to change this

pom.xml
```diff
@@ -89,23 +89,23 @@
         <maven.templating.version>1.0.0</maven.templating.version>
 
         <!-- dependencies -->
-        <jfx.version>18</jfx.version>
-        <kotlin.version>1.6.0</kotlin.version>
+        <jfx.version>18.0.2</jfx.version>
+        <kotlin.version>1.7.10</kotlin.version>
```
AlmasB, line 93: There are changes to multiple aspects of this file. Could we minimise the changes just to upgrading the kotlin and javafx version?

AlmasB (changes requested):
Thanks for this.

Happy to update if all builds pass, though comments should be addressed first please.

AlmasB:
merged as part of #1202

PR #1202, opened 2022-08-26
build: JavaFX 18 -> 19-ea+11, Kotlin 1.6.0 -> 1.7.10, attach 4.0.13 -> 4.0.15, jackson 2.12.1 -> 2.13.3

chengenzhao:
Hi:

We tried to bump the versions of dependencies and plugins
JavaFX from 18 -> 19-ea+10, it should be fine to change the final version to 19 since the JavaFX 19 is coming to the final release candidate, won't change too much
and JavaFX 19 has an important changes which is adding new .map and .flatMap method which would be really helpful for binding style programming

Kotlin from 1.6.0 -> 1.7.10
We have to remove the modular classifier of dependency of Kotlin-stdlib
since the incremental compilation is not experimental any more
and also another significant change in the code is that
@Suppress("NON_EXHAUSTIVE_WHEN") is not available since 1.7
non exhaustive when statement will cause compilation error
so use else clause to complete the when statement

and also bump the plugins' versions

The bumped version tests have passed in my computer
截屏2022-08-27 上午3 51 51

Cheers

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
@@ -158,7 +158,7 @@ data class MouseTrigger
                 KeyCode.CONTROL -> modifier == InputModifier.CTRL
                 KeyCode.SHIFT -> modifier == InputModifier.SHIFT
                 KeyCode.ALT -> modifier == InputModifier.ALT
-                else -> false
+                else -> isTriggered(event)
             }
         }
 
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
@@ -158,7 +158,7 @@ data class MouseTrigger
                 KeyCode.CONTROL -> modifier == InputModifier.CTRL
                 KeyCode.SHIFT -> modifier == InputModifier.SHIFT
                 KeyCode.ALT -> modifier == InputModifier.ALT
-                else -> modifier == InputModifier.NONE
+                else -> false
             }
         }
 
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
@@ -152,13 +152,13 @@ data class MouseTrigger
         return event.button == button && modifier.isTriggered(event)
     }
 
-    @Suppress("NON_EXHAUSTIVE_WHEN")
     override fun isReleased(event: InputEvent): Boolean {
         if (event is KeyEvent) {
-            when (event.code) {
-                KeyCode.CONTROL -> return modifier == InputModifier.CTRL
-                KeyCode.SHIFT -> return modifier == InputModifier.SHIFT
-                KeyCode.ALT -> return modifier == InputModifier.ALT
+            return when (event.code) {
+                KeyCode.CONTROL -> modifier == InputModifier.CTRL
+                KeyCode.SHIFT -> modifier == InputModifier.SHIFT
+                KeyCode.ALT -> modifier == InputModifier.ALT
+                else -> modifier == InputModifier.NONE
             }
         }
 
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/dev/Console.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/dev/Console.kt
@@ -92,10 +92,10 @@ class Console : Pane() {
 
             // add any special behaving keys
             setOnKeyPressed {
-                @Suppress("NON_EXHAUSTIVE_WHEN")
                 when (it.code) {
                     KeyCode.UP -> prev()
                     KeyCode.DOWN -> next()
+                    else -> Unit
                 }
             }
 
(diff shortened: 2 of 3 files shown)


fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
```diff
@@ -152,13 +152,13 @@ data class MouseTrigger
...
         if (event is KeyEvent) {
-            when (event.code) {
-                KeyCode.CONTROL -> return modifier == InputModifier.CTRL
-                KeyCode.SHIFT -> return modifier == InputModifier.SHIFT
-                KeyCode.ALT -> return modifier == InputModifier.ALT
+            return when (event.code) {
+                KeyCode.CONTROL -> modifier == InputModifier.CTRL
+                KeyCode.SHIFT -> modifier == InputModifier.SHIFT
+                KeyCode.ALT -> modifier == InputModifier.ALT
+                else -> false
```
chengenzhao: Although this code has passed the test
but still not sure about the return value of else
@AlmasB would you mind checking this line of code?
Is it return isTriggered(event) or false?

fxgl/src/main/kotlin/com/almasb/fxgl/dev/Console.kt
```diff
@@ -92,10 +92,10 @@ class Console : Pane() {
 
             // add any special behaving keys
             setOnKeyPressed {
-                @Suppress("NON_EXHAUSTIVE_WHEN")
                 when (it.code) {
                     KeyCode.UP -> prev()
                     KeyCode.DOWN -> next()
+                    else -> Unit
```
AlmasB, line 98: Fine with this, the `dev` package needs redesigning anyway

pom.xml
```diff
@@ -75,37 +75,37 @@
         <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
 
         <!-- plugins -->
-        <maven.gpg.version>1.6</maven.gpg.version>
-        <sonatype.nexus.staging.version>1.6.8</sonatype.nexus.staging.version>
-        <maven.compiler.version>3.8.0</maven.compiler.version>
+        <maven.gpg.version>3.0.1</maven.gpg.version>
```
AlmasB: Could we revert all plugins back to their original versions (unless they are required by the dependencies) so we can separate any potential implications.

pom.xml
```diff
@@ -75,37 +75,37 @@
...
+        <jfx.version>19-ea+10</jfx.version>
+        <kotlin.version>1.7.10</kotlin.version>
+        <attach.version>4.0.15</attach.version>
+        <jackson.version>2.13.3</jackson.version>
 
         <!-- test dependencies -->
-        <junit.jupiter.version>5.8.2</junit.jupiter.version>
-        <junit.platform.version>1.8.2</junit.platform.version>
-        <hamcrest.version>1.3</hamcrest.version>
+        <junit.jupiter.version>5.9.0</junit.jupiter.version>
```
AlmasB: Could we revert test deps back too, can handle them in a separate commit.

pom.xml
```diff
@@ -75,37 +75,37 @@
...
+        <hamcrest.version>2.2</hamcrest.version>
     </properties>
 
     <dependencies>
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>
             <artifactId>kotlin-stdlib</artifactId>
             <version>${kotlin.version}</version>
-            <classifier>modular</classifier>
+<!--            <classifier>modular</classifier>-->
```
AlmasB: Given there is no classifier now, we can delete this line.

AlmasB (commented):
Thanks for this, please can you have a look at the comments. Meanwhile, I'll check the logic behind `isReleased()` and get back.

pom.xml
chengenzhao: reverted
except sonatype.nexus.staging.version has to be bumped because the kotlin dependencies require higher version of this plugin

pom.xml
chengenzhao: reverted

pom.xml
chengenzhao: deleted

chengenzhao:
> Thanks for this, please can you have a look at the comments. Meanwhile, I'll check the logic behind `isReleased()` and get back.

Sure, un necessary bumped plugins and test dependencies are reverted

fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
AlmasB: Originally, `else` would just not run, falling through till the end, so I think it's `return isTriggered(event)`.

AlmasB:
Looks good, ready for merge once `isReleased()` is updated.

fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
chengenzhao:
```suggestion
                else -> isTriggered(event)
```

chengenzhao:
> Looks good, ready for merge once `isReleased()` is updated.

isReleased() is updated

but we got another problem which may be caused by jacoco plugin version 0.8.7
seems like jacoco 0.8.7 doesn't support class file format class file major version 62 which is Java 18 class file
It may cause the test of fxgl-io failed if maven uses JDK 18 to build the FXGL projects
So should we upgrade jacoco plugin version also?
Bumping the jacoco plugin version to 0.8.8 and the CI/CD passed all tests

Cheers

fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
chengenzhao: Done

AlmasB (approved):
looks good, many thanks

PR #1210, opened 2022-11-09
build: Kotlin 1.7.10 -> 1.8.0, Jackson 2.13.4 -> 2.14.1, github-action checkout/setup-java/cache v1,v2 -> v3, Attach 4.0.15 -> 4.0.16

chengenzhao:
Hi:

Today, Kotlin and Jackson release new version jars
and Jackson 2.13.4 has one [vulnerability](https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-databind/2.13.4) so we should better to upgrade it asap

for github action,
github action is deprecating checkout/setup-java/cache v1 & v2
so it has to upgrade to v3 otherwise the github action may not working in near future

Cheers

### DIFF
--- a/.github/workflows/deploy_SNAPSHOT.yml
+++ b/.github/workflows/deploy_SNAPSHOT.yml
@@ -10,10 +10,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Set up JDK 17
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v3
       with:
         java-version: 17
         server-id: ossrh
--- a/.github/workflows/deploy_SNAPSHOT.yml
+++ b/.github/workflows/deploy_SNAPSHOT.yml
@@ -15,6 +15,7 @@ jobs:
     - name: Set up JDK 17
       uses: actions/setup-java@v3
       with:
+        distribution: 'temurin'
         java-version: 17
         server-id: ossrh
         server-username: MAVEN_USERNAME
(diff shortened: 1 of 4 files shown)


chengenzhao:
The testing in MacOS failed
and I tried to run all tests in my computer (M1 mini) and all tests passed
so it could be the problems with the environment of Github action
the refresh rate is too low 
I think it is because the weak ability of intel chip Macos machine

截屏2022-11-09 17 40 52

.github/workflows/deploy_SNAPSHOT.yml
```diff
@@ -10,11 +10,12 @@ jobs:
...
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Set up JDK 17
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v3
       with:
+        distribution: 'temurin'
```
AlmasB, line 18: @chengenzhao Thanks for this. Looks good, just one question: why do we need to specify `distribution`

chengenzhao: Because this field is required when upgrading to Version 3
it could be ignored for v1&2, but is required when using v3 
截屏2023-01-04 上午11 17 33
https://github.com/actions/setup-java

chengenzhao: There are other options though
Oracle official provide its own JDK setup
e.g.
```github action
         uses: oracle-actions/setup-java@v1
         with:
           website: jdk.java.net
           release: 19
```
and Graal also provide JDK setup
e.g.
```github action
        uses: graalvm/setup-graalvm@v1
        with:
          version: 'latest'
          java-version: '19'
```
Both works

AlmasB: Great thanks

AlmasB (approved):

PR #1217, opened 2022-12-10
Add "bobbleDown" animation for Entities.

armlesshobo:


fxgl-core/src/main/kotlin/com/almasb/fxgl/animation/AnimationBuilder.kt
```diff
@@ -182,6 +182,13 @@ open class AnimationBuilder
             .from(Point2D(node.translateX, node.translateY))
             .to(Point2D(node.translateX, node.translateY + 5.0))
 
+    fun bobbleDown(entity: Animatable) = duration(Duration.seconds(0.15))
```
AlmasB, line 185: I think we can use this new implementation of `bobbleDown()` as a delegate for the `Node` one if we:

1. rename `entity -> obj`
2. delete implementation body of `bobbleDown(Node)`
3. make `bobbleDown(Node)` call this new `bobbleDown(Animatable)` by calling `node.toAnimatable()`

What do you think?

AlmasB (commented):
Thanks for this, please see below.

fxgl-core/src/main/kotlin/com/almasb/fxgl/animation/AnimationBuilder.kt
armlesshobo: That works for me. I'll make those changes. Thank you!

PR #1223, opened 2023-01-21
build: JavaFX 19 -> 19.0.2.1

chengenzhao:
There is a bug which may cause JavaFX on Mac crash & JavaFX 19.0.2.1 fix this bug
see also:
https://github.com/openjdk/jfx/pull/1003


AlmasB:
Many thanks.

asddcctrr143 (commented):
`<!DOCTYPE html>
<html>
`

PR #1232, opened 2023-02-23
Add Sample Titles

Harvalove:
Hi,
This PR adds titles to all the samples.


AlmasB:
Thanks for this, but for the sake of code simplicity, there is no need to provide a title for each sample.

PR #1244, opened 2023-03-05
Set/get method of current frame and additional methods of play(from & to)/loop(from) specific frame

### LINKED ISSUE #1198: Get current frame data & index in animations
### Discussed in https://github.com/AlmasB/FXGL/discussions/1191

<div type='discussions-op-text'>

<sup>Originally posted by **chengenzhao** July 30, 2022</sup>
Hi:

Is it possible to get the current frame data like which frame(1st, 2nd or 3rd) is currently showing
in the onUpdate method?
We tried to use method like viewComponent.getAnimationChannel().getCurrentFrame() but failed

Thanks</div>

### LINKED ISSUE #1243: Allow public API for getting and setting current frame data & index?
### Discussed in https://github.com/AlmasB/FXGL/discussions/1191

<div type='discussions-op-text'>

<sup>Originally posted by **chengenzhao** July 30, 2022</sup>
Hi:

Is it possible to get the current frame data like which frame(1st, 2nd or 3rd) is currently showing
in the onUpdate method?
We tried to use method like viewComponent.getAnimationChannel().getCurrentFrame() but failed

Thanks</div>

chengenzhao:
fix #1243 
fix #1198 
several new methods
```java
1. set/getCurrentFrame
2. animate() //play from current frame and stops on the last frame
3. playFrom(int startFrame) //same as setCurrentFrame(startFrame); animate();
4. loopFrom(int startFrame) //starts animation from startFrame and then keep looping
5. playTo(int pauseFrame) //start playing animation and pause on the specific frame
6. play(int startFrame, int pauseFrame) //star playing from startFrame and pause on the specific frame
8. playAnimationChannel(channel, startFrame) //same as setCurrentFrame(startFrame); continueAnimationChannel();
9. playAnimationChannel(channel, startFrame, pauseFrame) //play specific anime channel from start frame to pause frame
10. loopAnimationChannel(channel, startFrame) //loop animations but starts from specific frame
11. isPlaying and isPaused methods of AnimatedTexture
12. unit tests
```

### DIFF
--- a/fxgl-gameplay/src/main/java/module-info.java
+++ b/fxgl-gameplay/src/main/java/module-info.java
@@ -11,6 +11,7 @@ module com.almasb.fxgl.gameplay {
     requires com.almasb.fxgl.core;
     requires com.almasb.fxgl.scene;
     requires javafx.controls;
+    requires javafx.media;
     requires com.fasterxml.jackson.annotation;
 
     exports com.almasb.fxgl.achievement;
--- a/fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/cutscene/CutsceneService.kt
+++ b/fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/cutscene/CutsceneService.kt
@@ -17,6 +17,7 @@ import com.almasb.fxgl.cutscene.dialogue.DialogueScene
 import com.almasb.fxgl.cutscene.dialogue.FunctionCallHandler
 import com.almasb.fxgl.logging.Logger
 import com.almasb.fxgl.scene.SceneService
+import javafx.scene.media.MediaView
 
 /**
  *
@@ -31,7 +32,8 @@ class CutsceneService : EngineService() {
     private var gameVars: PropertyMap? = null
 
     private val scene by lazy { CutsceneScene(sceneService) }
-    val dialogueScene by lazy { DialogueScene(sceneService) }
+    private val dialogueScene by lazy { DialogueScene(sceneService) }
+    private val videoScene by lazy { VideoScene(sceneService) }
 
     @JvmOverloads fun startCutscene(cutscene: Cutscene, onFinished: Runnable = EmptyRunnable) {
         scene.assetLoader = assetLoader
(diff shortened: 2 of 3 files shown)


AlmasB:
Thanks for this, I should have time within the next several days to take a look at it. It's doing a bit more than just exposing current frame, so I might merge these incrementally.

AlmasB:
@chengenzhao it would be good to add you to the list of [Coordinators in README](https://github.com/AlmasB/FXGL/blob/dev/CONTRIBUTING.md#coordinator), if you are interested?

chengenzhao:
> @chengenzhao it would be good to add you to the list of [Coordinators in README](https://github.com/AlmasB/FXGL/blob/dev/CONTRIBUTING.md#coordinator), if you are interested?

Sure, that would be nice for me, thank you

chengenzhao:
> @chengenzhao it would be good to add you to the list of [Coordinators in README](https://github.com/AlmasB/FXGL/blob/dev/CONTRIBUTING.md#coordinator), if you are interested?

https://docs.oracle.com/javase/9/docs/api/javafx/animation/Animation.html#playFrom-java.lang.String-

I just read the Timeline class of JavaFX
Is it a good idea if we keep the same name as Timeline?
like jumpTo(), play() and playFromStart()
and that means we need to change the play method implementation 
the original play method will be renamed to playFromStart()
but because the jumpTo() method is brand new
so legacy code won't be affected

chengenzhao:
> Thanks for this, I should have time within the next several days to take a look at it. It's doing a bit more than just exposing current frame, so I might merge these incrementally.

https://docs.oracle.com/javase/9/docs/api/javafx/animation/Animation.html#playFrom-java.lang.String-

I just read the Timeline class of JavaFX
Is it a good idea if we keep the same name as Timeline?
like jumpTo(), play() and playFromStart()
and that means we need to change the play method implementation
the original play method will be renamed to playFromStart()
the new play method means play from that point(jumpTo & play)
but because the jumpTo() method is brand new
so legacy code won't be affected

PR #1257, opened 2023-03-28
fix: if title is empty, set to Untitled, closes #1225

### LINKED ISSUE #1225: When I set the title to null,it reports error
protected void initSettings(GameSettings gameSettings) {
        gameSettings.setTitle("");
    }
image

H-jj-R:


AlmasB:
Looks good, thanks!

PR #1258, opened 2023-03-28
test: added matrix 3x3 test

Mageconvict:


Mageconvict (commented):
Window header UI text scaling needs to be implemented in: src/main/kotlin/com/almasb/fxgl/ui/MDIWindow.kt

At line: 163.

AlmasB (approved):
Looks good, thanks!

PR #1286, opened 2023-08-04
feat: added a particle callback function that will be executed when the particle is removed

NexaCore168:
I think this function is still necessary.


AlmasB:
Hi, thanks for this. Please can you provide a use case example.

NexaCore168:
> Hi, thanks for this. Please can you provide a use case example.

https://www.bilibili.com/video/BV1iz4y1s7WW/?spm_id_from=333.337.search-card.all.click&vd_source=33227c9937f8f5c6629d4a8cb0e5be97
The URL above is a small demo. The raindrops in the video are simulated using a particle system. When raindrops fall onto the surface of the lake, they can cause ripples.

DeathPhoenix22:
Hi,
I was thinking of doing this! If it gets in the Engine, thanks for it :)

AlmasB:
@wonderfulLand Thanks for clarifying. I've just made a few editorial changes to maintain consistency. It's all merged now.

PR #1287, opened 2023-08-08
New methods play & playReverse of Animation class

chengenzhao:
Hi:

The animation class may need a play from current position method similar to the  same name method of JavaFX Animation class.
And this will simplify the further development of AnimatedTextures.

#1244


fxgl-core/src/main/kotlin/com/almasb/fxgl/animation/Animation.kt
```diff
@@ -66,6 +66,23 @@ abstract class Animation<T>(
...
+        if (!isAnimating) {
+            isReverse = true
+            play()
+        }
+    }
+
+    /**
+     * Plays Animation from current position in the direction indicated by rate.
+     */
+    fun play(){
```
AlmasB, line 79: I'm just wondering when this would be called. The current position in the animation can be controlled with `pause/resume`. `isAnimating` is always true when the animation is active. So the only way this is actually run is when the animation is fully stopped, meaning it matches the implementation of `start`. Does that make sense?

chengenzhao: let's say there is an AnimationChannel of wall clock starts from 0 to 12 clock
in the app, we want the clock animation starts from current time
so it would be nice to have
```kotlin
setTimeTo(currentTime)
play()
```
and later we may have
```kotlin
play(currentTime)//starts from currentTime
//or to more specific
playFrom(currentTime)
```
to replace the pure start() method
and later we may need playTo(specificPauseTime) method
and may have fluent method like
```java
animation.setTimeTo(currentTime)
                .playTo(specificTime) //when it comes to specific time, it will pause
                .playReverseTo(anotherTimeEarlier) //like resume to another time
                .play(); //continue playing the rest animation
//or
animation.play(from: currentTime, to: specificTime) //play from current time to specific time, when it comes to specific time, animation will pause
                .playReverseTo(anotherTimeEarlier) //like resume to another time
                .play(); //continue playing the rest animation
```

Issue #1297, opened 2023-10-10
The LoopRunner fpsBuffer2sec predictions can cause severe latency


DeathPhoenix22:
**Describe the bug**
The LoopRunner has a buffer strategy that will "predict" the next 2 seconds TPF based on the last 2 seconds registered.
This causes severe FPS drops when a single frame takes a long time to process as this exception is used to calculate the next TPF to use.

This enforces a 2 seconds lag period even though there is no lag anymore.

Note: If lag appears, the use of "scheduleAtFixedRate" will trigger all "ticks" even if their 0 millis long. In case of severe lag, it will put a lot of pressure on the game loop by causing many more calls, which is counter productive (causing further lag). Here is a brief example: If the scheduleAtFixedRate request 60 calls per seconds, and the first on update takes 1 second, the executor will call 59 ticks in a row just after that since it's 59 ticks late. In most cases, this freezes the game until the main loop is able to process all missed ticks fast enough so it can resync with the executor.

**To Reproduce**
1. An example will be provided along with a Unit Test. 
2. The test will intentionally perform significant lag (thread.wait) in 1 onUpdate. 

**Expected behavior**
The LoopRunner should always perform as many game Ticks as it can without intentionally slowing down a game. The provided solution will call the onUpdate based on the time between now and the lastFrame. Also, it'll skip "empty calls" from the "scheduleAtFixedRate" loop.

**Runtime info(please complete the following information):**
 - Windows 10/11
 - MacOSX Ventura

PR #1298, opened 2023-10-10
fix: Use effective tpf value in LoopRunner (onUpdate)

DeathPhoenix22:
Hi,

Issue Link: #1297 

This PR contains fixes for the FPS & the TPF calculations. The FPS calculation is now refreshed every 500 millis. The TPF will be based on current TPF.

Note that this change was a game changer. I was experiencing frequent freeze and FPS issues (I was wondering also why the Profiler was showing 60 steady FPS when it was lagging). With this change, the Profiler will show realtime FPS and recover from lag instantly (when possible).

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
@@ -77,6 +77,7 @@ internal class LoopRunner(
     fun resume() {
         log.debug("Resuming loop")
 
+        lastFrameNanos = 0
         impl.resume()
     }
 
@@ -93,15 +94,16 @@ internal class LoopRunner(
     }
 
     private fun frame(now: Long) {
+        val ticksPerSecond = if (ticksPerSecond < 0) 60 else ticksPerSecond // For JavaFX loops, cap at 60fps too
+
         if (lastFrameNanos == 0L) {
-            lastFrameNanos = now
-            lastFPSUpdateNanos = now
-            fpsSamplingCount = 0
+            lastFrameNanos = now - (1_000_000_000.0 / ticksPerSecond).toLong()
+            lastFPSUpdateNanos = lastFrameNanos
+            fpsSamplingCount = 1
         }
 
         tpf = (now - lastFrameNanos).toDouble() / 1_000_000_000
 
-        val ticksPerSecond = if (ticksPerSecond < 0) 60 else ticksPerSecond // For JavaFX loops, cap at 60fps too
         // The "executor" will call 60 times per seconds even if the game runs under 60 fps.
         // If it's not even "half" a tick long, skip
         if(tpf < (1_000_000_000 / (ticksPerSecond * 1.5)) / 1_000_000_000 ) {
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
@@ -46,6 +46,8 @@ internal class LoopRunner(
     var cpuNanoTime = 0L
         private set
 
+    private var lastFPSUpdateNanos = 0L
+    private var fpsSamplingCount = 0
     private var lastFrameNanos = 0L
 
     private val impl by lazy {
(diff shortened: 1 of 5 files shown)


fxgl/src/test/kotlin/com/almasb/fxgl/app/LoopRunnerTest.kt
```diff
@@ -106,44 +106,89 @@ class LoopRunnerTest {
...
-        assertThat(count2, lessThan(0.75))
+    @Test
+    @EnabledIfEnvironmentVariable(named = "CI", matches = "true")
+    fun `Lag Recovery`() {
+        var t = 0.0
+        var lag = 200L
+
+        listOf(
+            // run with a given ticks per second (via scheduled service tick)
+            LoopRunner(60) { t += it; Thread.sleep(lag) }
```
DeathPhoenix22, line 156: Please note that this test case doesn't use: LoopRunner { t += it; Thread.sleep(lag) } case. 
Mainly, JavaFX will not be affected by the Lag and will process 60 frames anyway.

fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
```diff
@@ -94,29 +93,35 @@ internal class LoopRunner(
...
             lastFPSUpdateNanos = now
-            fpsBuffer2sec = 0
+            fpsSamplingCount = 0
         }
 
-        cpuNanoTime = measureNanoTime {
-            runnable(tpf)
+        tpf = (now - lastFrameNanos).toDouble() / 1_000_000_000
+
+        val ticksPerSecond = if (ticksPerSecond < 0) 60 else ticksPerSecond // For JavaFX loops, cap at 60fps too
```
DeathPhoenix22: Please note that this is a safety mesure. I was supprised to see that JavaFX was defaulted to 120FPS on a Mac (instead of the 60FPS defined in the JavaFX documentation). In order to make tests "predictable", I had to cap it to 60FPS as each LoopRunner { t += it; Thread.sleep(lag) } test cases were failing by returning 120FPS

DeathPhoenix22:
Hi @AlmasB,

I've resolved a bug where pausing would cumulate all tpf while paused. I've adjusted the loop and the tests to reflect that. 

Sorry for that!

DeathPhoenix22:
Hi @AlmasB ,

I've been using this fix for quite a while now. I've since experienced stable FPS. I must say that this change is a game changer for the Engine. The previous LoopRunner was causing "fake" lag (highly noticeable whenever it happened).
I must admit that my project requires serious resources (RAM, CPU, GPU, HDD, etc), so I was facing the "fake lag" frequently.
I can't wait to show my project to the FXGL community, but it's not ready yet. (It's an RTS)

DeathPhoenix22:
@AlmasB Should I close my PRs?

DeathPhoenix22:
Without this PR, my game wouldn't run. The more resources your game needs, the worst the FPS gets without this fix. 
Also, the FPS indicated in the Profiling tool aren't accurate.

AlmasB:
Noted, I'll take a look at it over the weekend.

fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
```diff
@@ -94,29 +94,36 @@ internal class LoopRunner(
     }
 
     private fun frame(now: Long) {
-        if (lastFPSUpdateNanos == 0L) {
-            lastFPSUpdateNanos = now
-            fpsBuffer2sec = 0
+        val ticksPerSecond = if (ticksPerSecond < 0) 60 else ticksPerSecond // For JavaFX loops, cap at 60fps too
```
AlmasB: When ticksPerSecond is -1 or 0, JavaFX will match the display refresh rate (which is what we want), so we shouldn't cap it at 60

fxgl/src/main/kotlin/com/almasb/fxgl/app/LoopRunner.kt
```diff
@@ -94,29 +94,36 @@ internal class LoopRunner(
...
         }
 
-        fpsBuffer2sec++
+        fpsSamplingCount++
 
-        // if 2 seconds have passed
-        if (now - lastFPSUpdateNanos >= 2_000_000_000) {
+        // Update the FPS value every 500 millis
+        val timeSinceLastFPSUpdateNanos = now - lastFPSUpdateNanos;
+        if (timeSinceLastFPSUpdateNanos >= 500_000_000) {
```
AlmasB: I wonder if it's beneficial for the developers to have access to this "fps refresh interval", then we move the responsibility from us to them. We use a default value of X and let them pick a different one to suit their project.

AlmasB (changes requested):
Please see comments below

DeathPhoenix22:
Hi @AlmasB,

I've added the possibility to configure the FPS sampling rate.

Thanks,

AlmasB (approved):
I'll see if I can manage to uncap the 60 fps when ticks set to -1, since the contract is we match the display refresh rate. Everything else looks reasonable, many thanks.

Issue #1299, opened 2023-10-10
Cursor Texture corruption


DeathPhoenix22:
**Describe the bug**
When transitioning from a GameScene to the MainMenu, custom cursor texture gets corrupted.

**To Reproduce**
Steps to reproduce the behavior (tested with the TowerDefense example):
1. Add a custom cursor
2. Start a game (onGameUpdate)
3. Go back to MainMenu
4. The custom cursor texture gets corrupted

**Expected behavior**
Keep the custom cursor intact in all Scenes and SubScenes

**Screenshots**
Screenshots remove cursors. Sorry

**Runtime info(please complete the following information):**
 - Windows 10/11
 - Mac OSX Ventura

AlmasB:
@DeathPhoenix22 
Thanks for this, but by design every scene is separate, in case users wish to have different cursors for different scenes.

`getGameScene().setCursor()` only affects the game scene. The cursor in main / game menus can be set on an individual basis via their `root` or `contentRoot` when creating custom menus.

PR #1300, opened 2023-10-10
#1299

DeathPhoenix22:
Simple fix preventing Cursor corruption. The Dummy Scene isn't necessary at this stage since the currentScene is never null at that point. The Dummy Scene doesn't have the required information to drive the Cursor Texture, so we simply have to use the MainMenu.


DeathPhoenix22:
Will fix the commit history first. Sorry for that

PR #1301, opened 2023-10-10
fix: Prevent custom Cursor texture from getting corrupted when going back to the MainMenu

DeathPhoenix22:
Hi,

Issue Link: #1299 

By keeping the current scene intact (instead of switching to DummyScene), we prevent texture mixing.

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/FXGLApplication.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/FXGLApplication.kt
@@ -655,8 +655,6 @@ class FXGLApplication : Application() {
                 return
             }
 
-            // since mainMenu is a subscene we need an actual scene before it
-            mainWindow.setScene(dummyScene)
             mainWindow.pushState(mainMenu!!)
         }
 


DeathPhoenix22:
I'm closing as per @AlmasB comment, we need to set the cursor on every Scene.
Since scenes are not always accessible (Especially the private Dummy Scene of the menu) we need to do the following instead -> 
```
@Override
protected void onPreInit() {
    FXGL.getWindowService().getWindow$fxgl().getCurrentFXGLSceneProperty().addListener(((observableValue, previousScene, newScene) -> {
        newScene.setCursor(getAssetLoader().loadTexture("cursor/medium/cursor_default.png").getImage(), new Point2D(5.0, 5.0));
    }));
}
```
I'll probably write a State handler that will prevent "dumb" assignation of the Cursor when I reach that point in my Dev, but it works most of the time (Not always, there is a glitch between the MainMenu and the first Loading screen where the cursor goes for a millis back to the "OS" default cursor since this Listener is called too late in the thread chain) but it's minor.

AlmasB:
I think it is a good point that access to other scenes is convoluted. I'll have a think about a way to provide direct access to scenes.

Issue #1302, opened 2023-10-10
Support TMX texture relative Source path


DeathPhoenix22:
**Current**
The TilesetLoader is always looking for the Source texture path starting from the TMX map itself. This "forces" all textures to be placed next to the Maps & inside the levels folder. It basically consider only the Image name instead of the asset path.
**New**
The TilesetLoader will now take the complete path of the Source which allows us to place the resources anywhere relative to the file. For example, it's now possible to place the Tileset image file in the textures to avoid duplicating textures in both the textures folder and the levels folder. This also facilitate structure in the Levels folder (levels in different folders can share the same Source texture asset).
e.g. : <image source="../../textures/level/biome/ground.png"

DeathPhoenix22:
I've spent some time trying to provide a test, but the loadImage method of TilesetLoader fails silently. 


```
} catch (e: Exception) {
    log.warning("$imageName cannot be loaded using mapURL=$mapURL", e)

    resize(getDummyImage(), w, h)
}
```


Hopefully, the lack of automated testing is acceptable in this case? 
Thanks,

PR #1303, opened 2023-10-11
feat: Support TMX relative path for Source texture

DeathPhoenix22:
Hi,

Issue Link: #1302 

This change ensures that we can use the path specified on the tmx map file instead of removing the source file path at resource load time.


AlmasB:
Hi, could you add a test for this. Can we copy an existing texture into another test directory, so that the tmx references a relative path, that should make it fail (i.e. return a dummy texture)?

DeathPhoenix22:
Hi, 
I've added tests. It took me several hours to figure out a way to find a difference between a Dummy image and an Image. I finally decided to add-up all pixels. This is due to the fact that the dummy Image is a valid image with all correct properties :) 

I've added an extra test that validates the checksum strategy.

Hopefully, this suites the testing needs.

fxgl-entity/src/test/kotlin/com/almasb/fxgl/entity/level/tiled/TMXLevelLoaderTest.kt
```diff
@@ -361,6 +362,38 @@ class TMXLevelLoaderTest {
...
+        val levelDummyImage = TMXLevelLoader().load(javaClass.getResource("relative_image_source_missing.tmx"), GameWorld())
+        val levelDummyImageChecksum = imageChecksum((levelDummyImage.entities[0].viewComponent.children[0] as ImageView).image)
+        assertNotEquals(levelImageChecksum, levelDummyImageChecksum)
+    }
+
+    fun imageChecksum(image: Image): Int {
+        var sum = 0
+        for (x in 0 until image.getWidth().toInt()) {
+            for (y in 0 until image.getHeight().toInt()) {
+                sum += image.getPixelReader().getArgb(x, y)
```
AlmasB, line 391: I am not sure how big argb values are, so perhaps Long is more suitable. Though I suppose identical values will overflow in the same way, so Int should work too.

AlmasB (approved):
Looks great, many thanks.

Issue #938, opened 2021-01-22
Add onPreUpdate() (tentative name) that runs on bg thread before onUpdate


AlmasB:
Certain operations, such as AI, particle simulations, Canvas drawings, can hinder `onUpdate()` performance, thus leading to fps drop in JavaFX. To fix, we can run a single background tick before JavaFX `onUpdate()`.

This needs to be carefully designed, so discussion is welcome.

AlmasB:
One possible implementation is for `GameScene` to have a background thread and keep blocked with an `AtomicBoolean` while the JavaFX `onUpdate` runs. However, `onUpdate` still needs to wait for `onPreUpdate` to finish, which overall may not be a significant gain. Ultimately, it sounds like we are only gaining the time difference between JavaFX tick end and its next tick start.

It could be possible to run them in parallel, but `updateList` in `GameWorld` cannot have nested iterators.

DeathPhoenix22:
Hi Almas,

For my Game, I'll design a special type of service that will allow asynchronous (mainly parallelized) processing. It will basically allow having a Service running 1 tick behind allowing us to gain a full 18 millis extra processing (in 60FPS, for example.)

I've implemented a FogOfWar service that displays really efficiently, but in certain edge cases, it's requiring approximately 10 millis to resolve the situation, leaving only 8 millis for everything else. My goal is to put that process fully parallelized. The first onUpdate will trigger the processing with the known data and place it in a thread. On the next onUpdate call, the thread ".get()" (Future Promise) will hit FPS only if the parallelized treatment isn't finished yet.

Of course, this requires that the processing can run on an independent thread. For example, changing a view must be done in the main Thread (JavaFX limitation).

Hopefully, I'll provide a robust solution and create a PR for this soon.

AlmasB:
Hi @DeathPhoenix22 

Sounds like a good start. Looking forward to seeing the end result.

DeathPhoenix22:
Hi @AlmasB 
I've submitted an AsyncService implementation. If accepted, it would most likely require a small place in the official doc. 
In my case, my plan is to make my PathfindingService async & the AI too. The AI will most likely have a special "Offsync" service that will resync only when ready.

PR #1305, opened 2023-10-11
feat: AsyncService for parallel EngineService processing

DeathPhoenix22:
Hi,

Issue Link: #938

Here is an AsyncService with tests that will process the update in it's own thread. It will then resync on next onGameUpdate (Using standard Thread Promise strategy).

I've tested it on my own game and it was highly conclusive.

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/core/AsyncService.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/core/AsyncService.kt
@@ -1,0 +1,41 @@
+/*
+ * FXGL - JavaFX Game Library. The MIT License (MIT).
+ * Copyright (c) AlmasB (almaslvl@gmail.com).
+ * See LICENSE for details.
+ */
+
+package com.almasb.fxgl.core
+
+import java.util.concurrent.CompletableFuture
+
+/**
+ *
+ * @author Jean-Rene Lavoie (jeanrlavoie@gmail.com)
+ */
+abstract class AsyncService<T> : EngineService() {
+
+    private var asyncTask: CompletableFuture<T>? = null
+
+    /**
+     * Call the async game update. On next game update, wait for the task to be completed (if not already) and
+     * call onPostGameUpdateAsync to allow JavaFX thread dependent task handling (e.g. updating the Nodes)
+     */
+    override fun onGameUpdate(tpf: Double) {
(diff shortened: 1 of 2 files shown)


DeathPhoenix22:
I'm still using it and it brings Multi-Threading to the Framework. For highly intensive games, this is required.

AlmasB:
This looks fine, could I have a look at what the call site looks like where you are using it? I'm wondering if this should be a different type of concept, rather than service.

Is this roughly how it is going to be used?

```
class MyAsyncService extends AsyncService<Type>

settings.addEngineService(MyAsyncService.class);
```

DeathPhoenix22:
@AlmasB Gladly, here is a concrete case where I use it:

My FogOfWarService is currently processing updates async to avoid running on the "main" game thread. This grants 1 extra CPU to process any service tasks (that can be process Async, such as States updates, effects processing, etc). 

As you mentioned, you can simply declare an async service this way:
```class FogOfWarService extends AsyncService<Void>```

It's also interesting to know that an "Async" service is also synched, meaning that you still have access to onGameUpdate. For example, my FogOfWarService implements onGameUpdate & the 2 new methods:
```
    public Void onGameUpdateAsync(double tpf) {
        fogLayer.onUpdateAsync();
        return null;
    }

    @Override
    public void onPostGameUpdateAsync(Void result) {
        fogLayer.onPostUpdateAsync();
    }
```

Here is the simple "flow": 
1. on tick 1 
    a. onPostUpdateAsync() is called to resync the result of the async update (in this case, there is nothing to sync)
    b. onGameUpdateAsync(), but the Engine doesn't "wait" for the response allowing a full extra tick of parallel processing
2. on tick 2
    a. onPostUpdateAsync() is called to resync the result of the async update. This is really important as all JavaFX related operations must occur in the JavaFX thread!
   b. onGameUpdateAsync(), again preparing the result for tick 3

Also, I've included the Return type which could allow Result transfer directly from the Async processing back to the PostAsync processing. Let's say the Async method builds a Grid and the Async uses it afterwards. In my case, the fogLayer is concurrent safe already so it wasn't required.

Lastly, yes it's a normal Service so it can be added this way, as you mentioned:
```
settings.addEngineService(FogOfWarService.class);
```

AlmasB (approved):
Thanks for the use case, it is clear now.

AlmasB:
@DeathPhoenix22 FYI, it's now under `...core.concurrent.`

Issue #1310, opened 2023-10-24
Allow Volume control of Music & Sound with Property 


DeathPhoenix22:
**Volume manipulation with Property**
It's currently possible to change the volume of Music and Sound. The fact that we don't have access to a volume property prevents the use of Timeline effects (FadeIn, FadeOut for example). 

Granting access to the Volume property of the native Javafx Audio entity would bypass the current Global Volume implementation. The solution most allow both features to coexist.

**Work Around**
```
Music music = getAssetLoader().loadMusic("rain.mp3");
music.getAudio().setVolume(0.0);
DoubleProperty volumeHook = new SimpleDoubleProperty(0.0);
volumeHook.addListener((observableValue, number, t1) -> {
    music.getAudio().setVolume((double)t1);
});
Timeline timeline = new Timeline(new KeyFrame(Duration.seconds(5), new KeyValue(volumeHook, 1.0)));
FXGL.getAudioPlayer().playMusic(music);
timeline.play();
```

**Real Life Example**
To provide additional context, I'm building an RTS. This means that sound may play even though you don't see the source, but it's distance from the Camera affects it's native volume. Also, the closer you are from the source (Zoom level), the higher the volume is. Another aspect of the sound effect system is ambient sound such as Rain. The rain doesn't appear directly from volume 0 to 1. It will fade in and fade out. You can also easily consider that the sound will get heavier the harder the storm is (I also have a Storm system which will require Rain to consider the intensity of the weather). 
I should normally also Pan the sound (Left/Right) based on the source location from the player screen, but this part will be left for another PR, if time permits.

CC: @AlmasB @chengenzhao

Issue #1187, opened 2022-07-20
Adding more options for volume


AlmasB:
### Discussed in https://github.com/AlmasB/FXGL/discussions/1180

<div type='discussions-op-text'>

<sup>Originally posted by **nicymike** June 10, 2022</sup>
Hey!

1. What do you think about adding a new setting to audio for master volume.
Many games have three (or sometimes more) volume sliders: master, music, sound.
The volume for playing audio is then calculated by:
soundVolumeForPlay = master * sound
musicVolumeForPlay = master * music

2. What do you think about adding the possibility to play a sound with a different volume eg. adding a new optional parameter to the methods for playing a sound or music.
In my game I have certain sounds for ambience which should be played with lower volume.
What I am currently doing with FXGL is: 1. Lower sound volume, 2. Play sound, 3. Raise sound volume.
</div>

AlmasB:
Relevant extract:

> That's a good point. It seems the implementation is hidden from public API. This might have been necessary to hide cross-platform support (particularly on mobile), but I don't see why we couldn't open access to getAudio(). From the list you will note the method is actually there, just hidden in $fxgl_core. I'm happy to turn this into an issue, so we can track it and fix it eventually.

PR #1311, opened 2023-10-24
feat: Allow volume control of Music & Sound with property

DeathPhoenix22:
Hi,

Issue Link: https://github.com/AlmasB/FXGL/issues/1310

Here's an implementation that allow custom Volume configuration of all Sounds & Musics. I've placed the "Master / Global" volume property of the AudioType inside them. This felt pretty natural since the "expert" (GRAPS) of that information is the AudioType itself. This also mean that we could really easily add a new AudioType without much efforts and that new type would automatically mix itself against the Global and the current Audio.

@AlmasB NOTE: I took the initiative to revert the changes made in #1187. Granting access to the internal audio is causing a bug related to volume adjustments in the settings. Whenever a player was adjusting his Global volume, all custom changes made on the audio (getAudio()) was automatically reverted. With this implementation, the conjunction of the 2 values are always kept (via the mix() method)

Example:
```
Music music = getAssetLoader().loadMusic("rain.mp3");
music.setVolume(0.0);
Timeline timeline = new Timeline(
    new KeyFrame(Duration.seconds(5),
    new KeyValue(music.volumeProperty(), 1.0))
);
FXGL.getAudioPlayer().playMusic(music);
timeline.play();
```

Note that changing the Global Music volume will also automatically impact all Music:
```
AudioType.MUSIC.getVolume().set(AudioType.MUSIC.getVolume().get() - 0.1);
```

I've tested it on my own game and I needed it!
I hope you like it! :)

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
@@ -11,7 +11,6 @@ import com.almasb.fxgl.core.EngineService
 import com.almasb.fxgl.core.Inject
 import com.almasb.fxgl.core.collection.UnorderedArray
 import com.almasb.fxgl.logging.Logger
-import javafx.beans.property.DoubleProperty
 import java.net.URL
 
 /**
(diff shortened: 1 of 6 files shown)


fxgl-core/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
```diff
@@ -27,27 +26,11 @@ class AudioPlayer : EngineService() {
...
-
-    @Inject("globalSoundVolumeProperty")
-    private lateinit var soundVolume: DoubleProperty
-
     @Inject("isPauseMusicWhenMinimized")
     private var isPauseMusicWhenMinimized = true
 
     private val loader = DesktopAndMobileAudioLoader()
 
-    override fun onMainLoopStarting() {
```
DeathPhoenix22, line 41: Automatically handled by the Audio with Property change listeners

fxgl-core/src/main/kotlin/com/almasb/fxgl/audio/AudioPlayer.kt
```diff
@@ -74,7 +57,6 @@ class AudioPlayer : EngineService() {
         if (!activeSounds.containsByIdentity(sound))
             activeSounds.add(sound)
 
-        sound.audio.setVolume(soundVolume.value)
```
DeathPhoenix22, line 77: No need anymore, the global volumes are automatically mixed with the Media configured sound

Issue #1318, opened 2023-11-06
Particules Emitters are Spawning particules at zIndex = 0


DeathPhoenix22:
**Describe the bug**
I've noticed that all particules were draw under anything that had a zIndex set.

**To Reproduce**
1. Create a Rectangle with zIndex = 1000 that fills the screen
2. Create a Particules Emitter (I tried Rain & Smoke, same behavior)
3. Start the emitter
4. You won't see any particules

**Expected behavior**
Particules need to be displayed at a desired zIndex.

**Additional context / possible fixes**
The Bug is caused by this strategy:
```
open class ParticleComponent(val emitter: ParticleEmitter) : Component() {
    ...
    /**
     * This is the entity whose view is used to render particles.
     * Use of extra entity allows to render particles independently from
     * the entity to which this component is attached. Otherwise, the entire
     * set of emitted particles will be moved based on entity's view.
     */
    val parent = Entity()
    ...
```
In fact, the ParticuleComponent will create it's own "ghost" entity instead of using the Entity tied to the Component. This ensures that particules stay were they were emitted from (which is a desired behavior, of course). The Particule has an ImageView that will ensure that the particule is spawned at the X & Y desired positions. But, the problem is related to how the zIndex sorting works. An ImageView doesn't have a zIndex strategy, it's handled by the FXGL Scene (re-adding Nodes to the Scene after a zIndex sort).

**Solution**
Simply assign the zIndex of the real ParticuleComponent entity to the "Ghost" entity by default. Important: The "Ghost" entity zIndex will not be bind to the real ParticuleComponent entity, to prevent Particule relocations.

@AlmasB I'll create a PR soon.

PR #1319, opened 2023-11-06
fix: Emit particles at initial Entity zIndex instead of zIndex 0

DeathPhoenix22:
Hi @AlmasB,
Here is the PR for the Bug #1318 
Thanks,

### DIFF
--- a/fxgl-entity/src/test/kotlin/com/almasb/fxgl/particle/ParticleComponentTest.kt
+++ b/fxgl-entity/src/test/kotlin/com/almasb/fxgl/particle/ParticleComponentTest.kt
@@ -13,7 +13,6 @@ import org.hamcrest.CoreMatchers.`is`
 import org.hamcrest.MatcherAssert.assertThat
 import org.junit.jupiter.api.Assertions.assertNotNull
 import org.junit.jupiter.api.Assertions.assertNull
-import org.junit.jupiter.api.Assertions.assertNotNull
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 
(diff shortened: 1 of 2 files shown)


fxgl-entity/src/test/kotlin/com/almasb/fxgl/particle/ParticleComponentTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.jupiter.api.Test
+
+/**
+ *
+ * @author Jean-Rene Lavoie (jeanrlavoie@gmail.com)
+ */
+class ParticleComponentTest {
+
+    private lateinit var world: GameWorld
+    private lateinit var particule: ParticleComponent
```
AlmasB: `particule` -> `particle`

fxgl-entity/src/test/kotlin/com/almasb/fxgl/particle/ParticleComponentTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @BeforeEach
+    fun setUp() {
+        world = GameWorld()
+        particule = ParticleComponent(ParticleEmitter())
+    }
+
+    @Test
+    fun `Create ParticuleComponent with zIndex`() {
```
AlmasB: `Particule` -> `Particle`

fxgl-entity/src/test/kotlin/com/almasb/fxgl/particle/ParticleComponentTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        world.addEntity(e)
+        particule.onAdded()
+        particule.onUpdate(1.0)
+
+        assertThat(particule.parent.zIndex, `is`(100))
+
+        e.zIndex = 200
+        particule.onUpdate(1.0)
+
+        assertThat(particule.parent.zIndex, `is`(100))
```
AlmasB: This is more of a design question. By using the same logic, shouldn't this be 200 to match zIndex of `e`? If yes, then the implementation could be something like `parent.zIndex bind to entity.zIndex` using their property accessors. Does this make sense?

AlmasB (commented):
Thanks for this. Please can you take a look the comments when you get a chance.

DeathPhoenix22:
@AlmasB I've update the name. Thx for pointing this out

fxgl-entity/src/test/kotlin/com/almasb/fxgl/particle/ParticleComponentTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        world.addEntity(e)
+        particle.onAdded()
+        particle.onUpdate(1.0)
+
+        assertThat(particle.parent.zIndex, `is`(100))
+
+        e.zIndex = 200
+        particle.onUpdate(1.0)
+
+        assertThat(particle.parent.zIndex, `is`(100))
```
AlmasB, line 53: Shouldn't `particle.parent.zIndex` be 200 here, to match `e.zIndex`?

DeathPhoenix22: The particle inherits the initial Entity zIndex. The second test is to validate that the particle zIndex is then independent which allows mutating particles independently after creation (moving some over, others under). 
All particles are under the same parent, this allows full control over their display position after creation.

DeathPhoenix22: Example: You have a Rain Emitter. The emitter is the "source" / "origin" spawning point. You spawn droplets and reduce their independent zIndex to emulate their position in a Parallax, for example. This means that droplets are able to show over a tree, but then it could show under the tree once it's zIndex reached the "behind the tree" value. Or even better, smoke particles generated in the air can get behind objects on the ground gradually.

AlmasB: Thanks for the explanation. I quite like the idea of particles being under / over entities, though not sure we currently support that, given we only have 1 parent and its zIndex is fixed. Something for later discussion perhaps. As far as the issue is concerned, I think it's fixed.

AlmasB (approved):
Thanks for the fix.

PR #1334, opened 2024-01-19
feat: Support diagonal pathfinding movement

DeathPhoenix22:
Hi @AlmasB ,

I've completed the implementation to allow diagonal Pathfinding movement. This implementation is "backward compatible" by setting the parameter Diagonal to "Never" by default. To enable Diagonal movement, simply provide the Diagonal.ALLOWED flag in the constructor or through it's Setter.

I've defaulted the AStarPathfindingSample to Diagonal.ALLOWED.

As requested, I'll do the changes "step by step".
Note: Because I'm doing it "step by step", this means that future PRs will depend on each others.

Possible Upcoming changes (No promises):
- Optimizations -> Reduced memory allocation, Faster calculation, etc
- Include a "No Corner Crossing" option -> This is to prevent "noisy" collisions detection when an Entity walks close to a diagonal wall
- Flexible CellState definition -> Ability to define your own CellStates & allow defining "allowed" states while retrieving the Path. For example, you could wish to have a single state grid that will support Ground vs Water vs Amphibious pathfinding
- Allow AStarGrid (.fromWorld(...)) to be loaded by Entity instead of restricting it by entity types. Currently, only the type of the entity may be used to define if a Cell is walkable or not. Having the Entity directly offers more flexibility.
- 16 Directions Pathfinding -> Tricky, but it would add a noticeable "smoothness" to Entity movements

Let me know if I should adjust it.

### DIFF
--- a/fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
+++ b/fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
@@ -234,19 +234,19 @@ public class Grid<T extends Cell> {
     }
 
     public final Optional<T> getUpRight(int x, int y) {
-        return getOptional(x + 1, y + 1);
-    }
-
-    public final Optional<T> getUpLeft(int x, int y) {
-        return getOptional(x - 1, y + 1);
-    }
-
-    public final Optional<T> getDownRight(int x, int y) {
         return getOptional(x + 1, y - 1);
     }
 
-    public final Optional<T> getDownLeft(int x, int y) {
+    public final Optional<T> getUpLeft(int x, int y) {
         return getOptional(x - 1, y - 1);
+    }
+
+    public final Optional<T> getDownRight(int x, int y) {
+        return getOptional(x + 1, y + 1);
+    }
+
+    public final Optional<T> getDownLeft(int x, int y) {
+        return getOptional(x - 1, y + 1);
     }
 
     /**
--- a/fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
+++ b/fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
@@ -7,6 +7,7 @@
 package com.almasb.fxgl.core.collection.grid;
 
 import com.almasb.fxgl.core.math.FXGLMath;
+import static com.almasb.fxgl.core.collection.grid.Diagonal.*;
 
 import java.lang.reflect.Array;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
```diff
@@ -29,6 +29,7 @@ public class Grid<T extends Cell> {
     private final int height;
     private final int cellWidth;
     private final int cellHeight;
+    private Diagonal diagonal;
```
AlmasB, line 32: The grid in itself is just a data structure. I think the fact that diagonal movement is supported or not could be extracted out of this and dictated by higher level code/system.

fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
```diff
@@ -171,6 +201,22 @@ public final Optional<T> getDown(Cell cell) {
         return getDown(cell.getX(), cell.getY());
     }
 
+    public final Optional<T> getUpRight(Cell cell) {
```
AlmasB, line 204: These are useful convenience methods for the first of the two PRs.

fxgl-core/src/main/java/com/almasb/fxgl/core/collection/grid/Grid.java
```diff
@@ -136,6 +157,15 @@ public final List<T> getNeighbors(int x, int y) {
         getUp(x, y).ifPresent(result::add);
         getRight(x, y).ifPresent(result::add);
         getDown(x, y).ifPresent(result::add);
+
+        // Include "Corner" neighbors only if diagonal movement is allowed
```
AlmasB, line 161: This should go into a new method `getNeighbors` that allows us to provide info on how to extract the neighbors.

```
getNeighbors(x, y, NeighborFilteringOption.FOUR_DIRECTIONS)

NeighborFilteringOption.EIGHT_DIRECTIONS
```
If we need more in the future, we can modify as appropriate.

AlmasB (changes requested):
Thanks for this, it looks really good.

Could we separate it into two PRs, it will be easier to review and merge.

1) Introduce the concept of diagonal neighbors.
2) Implement A* support of diagonal movement.

I've left some comments related to (1). Once it is merged, we can start looking at (2).

DeathPhoenix22:
> Thanks for this, it looks really good.
> 
> Could we separate it into two PRs, it will be easier to review and merge.
> 
> 1. Introduce the concept of diagonal neighbors.
> 2. Implement A* support of diagonal movement.
> 
> I've left some comments related to (1). Once it is merged, we can start looking at (2).

Hi @AlmasB , from a functional perspective, both PRs will be tightly coupled and useless without the other. 
Failing to merge the second PR will render the 1st PR dead code. 

I'll comply, but I don't recommend segmenting this PR.
Have a nice day :)

PR #1339, opened 2024-01-31
feat: Support 8 directions AStar pathfinding

DeathPhoenix22:
Hi @AlmasB ,

As discussed, here is the second part of the 8 directions Pathfinding support. 
It includes the Heuristic value integration where it can be customized (both algorithm & weight).

Please let me. know if I need to adjust it.
Thanks,


AlmasB (approved):
Looks good, thanks!

PR #1340, opened 2024-02-12
Property map serializable type#1250

### LINKED ISSUE #1250: PropertyMap should have SerializableType
This allows the map to be easily written to and read from a Bundle

BeckFW:
Adding SerializableType to PropertyMap to close #1250. Also includes documentation and test case.

@artemijsem

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/core/collection/PropertyMap.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/core/collection/PropertyMap.kt
@@ -314,9 +314,9 @@ class PropertyMap : SerializableType {
 
     override fun write(bundle: Bundle) {
         // Convert to string map
-        this.toStringMap().forEach { key, value ->
+        this.toStringMap().forEach { (key, value) ->
             // write to bundle
-            bundle.put(key, value);
+            bundle.put(key, value)
         }
     }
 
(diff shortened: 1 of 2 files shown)


fxgl-core/src/main/kotlin/com/almasb/fxgl/core/collection/PropertyMap.kt
```diff
@@ -309,6 +312,20 @@ class PropertyMap {
...
+    override fun write(bundle: Bundle) {
+        // Convert to string map
+        this.toStringMap().forEach { key, value ->
+            // write to bundle
+            bundle.put(key, value);
+        }
+    }
+
+    override fun read(bundle: Bundle) {
+        bundle.data.forEach { (key) ->
```
AlmasB: bundle.data is a `Map<String, Serializable>`, so `forEach { key, value -> ... }` should be possible, in which case `bundle.get(key)` can be replaced by just `value` and avoid extra value lookups.

Please check the above is true and update if so.

AlmasB (commented):
Thanks, looks good, please check the comment

fxgl-core/src/main/kotlin/com/almasb/fxgl/core/collection/PropertyMap.kt
BeckFW: `toValue(bundle.get(key))` has been replaced by `toValue(value.toString())`. Using `value` directly caused tests to fail. Still avoids the extra lookup from before.

AlmasB:
Brilliant, thank you

AlmasB (approved):

PR #1342, opened 2024-02-12
feat: Software Cursor, closes #1090

### LINKED ISSUE #1090: Software cursor
Embedded devices may not have access to a mouse (or if using controllers only), hence a software cursor, with API that allows users to control it is beneficial. Leveraging the input mocking API already present (particularly related to mouse events), this feature could be a useful substitute when running without a mouse.

@DieterHolz, thanks for the idea. I think this could very well be a student project given it is possible to clearly define the requirements and the fact that the goal is tractable.

alexmoore04:
adds virtual cursor
@Leo-Waters
@popsx0x


AlmasB:
Thanks, please can you check why the PR is failing checks

AlmasB (approved):
Thanks for the base implementation

Issue #796, opened 2020-05-05
Fix trigger sequence mini game


AlmasB:
https://github.com/AlmasB/FXGL/blob/dev/fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/minigames/triggersequence/TriggerSequenceMiniGame.kt

At L96, 

https://github.com/AlmasB/FXGL/blob/43a1a24ac82b03ed733734c2c35ececb0bf5c4e2/fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/minigames/triggersequence/TriggerSequenceMiniGame.kt#L96

there should be an extra check for "*and the view is between line1 and line2*"

BeckFW:
@AlmasB Discussed and fixed. Please assign to
@artemijsem @dalehuntgb @beckfw

AlmasB:
Great, looking forward to the PR 😄

PR #1345, opened 2024-02-20
TriggerSequence Minigame: Check if trigger is between two lines #796

BeckFW:
Fix for issue #796 
Changes: 
- Move creation of line objects out of the constructor
- Check position of trigger vs position of both lines before triggering "good" animation

@AlmasB please review and let me know if further changes are needed

@dalehuntgb @artemijsem

### DIFF
--- a/fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/minigames/triggersequence/TriggerSequenceMiniGame.kt
+++ b/fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/minigames/triggersequence/TriggerSequenceMiniGame.kt
@@ -37,14 +37,16 @@ class TriggerSequenceView(miniGame: TriggerSequenceMiniGame = TriggerSequenceMin
     private val circle = StackPane()
     private val bg = Circle(40.0, 40.0, 40.0, Color.GREEN)
 
+    private val line1 = Line(0.0, 0.0, 0.0, 300.0)
+    private val line2 = Line(100.0, 0.0, 100.0, 300.0)
+
     private val triggerViews = Group()
 
     private val good = ImageView(Image(javaClass.getResourceAsStream("checkmark.png")))
     private val bad = ImageView(Image(javaClass.getResourceAsStream("cross.png")))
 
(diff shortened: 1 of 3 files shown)


fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/minigames/triggersequence/TriggerSequenceMiniGame.kt
```diff
@@ -87,13 +89,18 @@ class TriggerSequenceView(miniGame: TriggerSequenceMiniGame = TriggerSequenceMin
             override fun onActionBegin(trigger: Trigger) {
                 if (trigger is KeyTrigger) {
                     val key = trigger.key
+                    val currentTrigger = triggerViews.children[0]
```
AlmasB: Is this taking just the first trigger view? What about other ones?

fxgl-gameplay/src/main/kotlin/com/almasb/fxgl/minigames/triggersequence/TriggerSequenceMiniGame.kt
```diff
@@ -87,13 +89,18 @@ class TriggerSequenceView(miniGame: TriggerSequenceMiniGame = TriggerSequenceMin
...
 
                     circle.opacity = 1.0
 
                     if (triggerViews.children.isNotEmpty())
                         triggerViews.children.removeAt(0)
 
-                    if (miniGame.press(key)) {
+                    // Has the correct key been pressed, and is the current view between the two lines
+                    if (miniGame.press(key) &&
+                        currentTrigger.translateX >= line1.startX &&
```
AlmasB, line 134: We are potentially being too strict here from the gameplay perspective. What if the current view X is less than line1.x but the right side of the view is still between line1.x and line2.x? I think we should allow it to be accepted

AlmasB (commented):
Thanks for this, a few questions below

AlmasB (approved):
Looks good, thanks!

Issue #1021, opened 2021-04-21
FXGLDefaultMenu layout is not responsive


AlmasB:
module: `fxgl`
package: `com.almasb.fxgl.app.scene`
file: `FXGLDefaultMenu.kt`

https://github.com/AlmasB/FXGL/blob/43a1a24ac82b03ed733734c2c35ececb0bf5c4e2/fxgl/src/main/kotlin/com/almasb/fxgl/app/scene/FXGLDefaultMenu.kt#L90-L92

These are hardcoded, instead we should listen for (bind to) settings.prefWidth/Height properties.

You can check the result of this working (or not working) in `fxgl-samples` module by running a [MenuSample](https://github.com/AlmasB/FXGL/blob/dev/fxgl-samples/src/main/java/intermediate/MenuSample.java) and adding `settings.setManualResizeEnabled(true);`, then resizing the game at runtime.

Currently on resize, the hardcoded values will not result in a responsive layout.

DaleHuntGB:
Discussed & Fixed.

@artemijsem
@BeckFW

PR #1356, opened 2024-02-27
FXGLDefaultMenu layout is not responsive #1021

DaleHuntGB:
- Position UI elements appropriately depending on `appWidth` size, ensuring it is a little more aesthetic and consistent across multiple previews. 

This solves #1021.

Help from: @Artemijsem & @BeckFW


AlmasB:
Thanks, looks good, there's an extra file in this PR that shouldn't have changed -- `BuildProperties.java`. Please check the files changed and remove it

DaleHuntGB:
Oh yeah, it would not build unless I copied it accordingly (not sure why that was the issue), am I safe to delete it from the PR? I didn't think it would have pushed this changed as I only staged the `FXGLDefaultMenu`

AlmasB:
@DaleHuntGB 
Yes, it is an auto-generated file, which will be created during the build process. Safe to delete from the PR

DaleHuntGB:
> @DaleHuntGB Yes, it is an auto-generated file, which will be created during the build process. Safe to delete from the PR

`BuildProperties` has been deleted from PR (I hope).

PR #1359, opened 2024-03-03
Dungeon Generator

keziah480:
+ new DungeonCell class, extending Cell with dungeon-specific data
+ new Dungeon class, capable of generating simple connected dungeons with a variety of circular and rectangular rooms

Here is an example generation with hashes representing walls:

image


AlmasB (approved):
Looks great, thanks.

PR #1362, opened 2024-03-21
feat: Option to add child at First position to ViewComponent

DeathPhoenix22:
Hi @AlmasB,

Add the possibility to add child to ViewComponent Last vs First.

No change to default behavior.
Reference: #1361


fxgl-entity/src/main/kotlin/com/almasb/fxgl/entity/components/ViewComponent.kt
```diff
@@ -214,7 +221,11 @@ class ViewComponent : Component() {
             }
         }
 
-        group.children += child
+        if(addLast) {
```
AlmasB, line 224: `addLast` is currently being used to add groups to parent -- in-game groups go first, developer groups go last, so we cannot use it to manage group children.

Perhaps we can generalise this to something like `addChild(index: Int, node: Node, isTransformApplied: Boolean)` with the default implementation being just `addChild(node: Node, isTransformApplied: Boolean)` which adds it last.

AlmasB (commented):
Thanks, please see below

Issue #1020, opened 2021-04-21
FXGLDefaultMenu::createTitleView - incorrect layout width calculation


AlmasB:
```
        // the lazy font size may not have been computed by this point
        // so we need to listen for changes in layout bounds of text1 and text2

val textWidth = text.layoutBounds.width + text2.layoutBounds.width
```

https://github.com/AlmasB/FXGL/blob/43a1a24ac82b03ed733734c2c35ececb0bf5c4e2/fxgl/src/main/kotlin/com/almasb/fxgl/app/scene/FXGLDefaultMenu.kt#L199

PR #1365, opened 2024-03-26
fix: listen for changes in layout bounds in FXGLDefaultMenu.kt

Bart-del:
Hi!
Issue link: #1020 

This PR contains fix for title width calculation. Now it will listen for changes in layoutBounds and should update title width when font size will change. I've also changed the variable names to be more descriptive.

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/app/scene/FXGLDefaultMenu.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/app/scene/FXGLDefaultMenu.kt
@@ -193,20 +193,20 @@ open class FXGLDefaultMenu(type: MenuType) : FXGLMenu(type) {
         firstLetter.strokeProperty().bind(titleColor)
         firstLetter.strokeWidth = 1.5
 
-        val restOfTheTittle = getUIFactoryService().newText(title.substring(1, title.length), 50.0)
-        restOfTheTittle.fill = null
-        restOfTheTittle.stroke = titleColor.value
-        restOfTheTittle.strokeWidth = 1.5
+        val restOfTheTitle = getUIFactoryService().newText(title.substring(1, title.length), 50.0)
+        restOfTheTitle.fill = null
+        restOfTheTitle.stroke = titleColor.value
+        restOfTheTitle.strokeWidth = 1.5
 
         var firstLetterWidth = firstLetter.layoutBounds.width
-        var restOfTheTittleWidth = restOfTheTittle.layoutBounds.width
+        var restOfTheTitleWidth = restOfTheTitle.layoutBounds.width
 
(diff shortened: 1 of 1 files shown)


fxgl/src/main/kotlin/com/almasb/fxgl/app/scene/FXGLDefaultMenu.kt
```diff
@@ -188,19 +188,25 @@ open class FXGLDefaultMenu(type: MenuType) : FXGLMenu(type) {
...
+        var firstLetterWidth = firstLetter.layoutBounds.width
+        var restOfTheTitleWidth = restOfTheTitle.layoutBounds.width
 
-        val border = Rectangle(textWidth + 30, 65.0, null)
+        firstLetter.layoutBoundsProperty().addListener { _, _, newBounds -> firstLetterWidth = newBounds.width}
+        restOfTheTitle.layoutBoundsProperty().addListener { _, _, newBounds -> restOfTheTitleWidth = newBounds.width}
+
+        val titleWidth = firstLetterWidth + restOfTheTitleWidth
+
+        val border = Rectangle(titleWidth + 30, 65.0, null)
```
AlmasB, line 209: The width of this rectangle is still fixed, it should be updated based on changes to layout bounds of the text objects. Something like:

```
firstLetter.layoutBoundsProperty().addListener { _, _, newBounds -> update border width }
```

fxgl/src/main/kotlin/com/almasb/fxgl/app/scene/FXGLDefaultMenu.kt
```diff
@@ -224,11 +230,11 @@ open class FXGLDefaultMenu(type: MenuType) : FXGLMenu(type) {
         emitter.setSpawnPointFunction { Point2D.ZERO }
         emitter.setAccelerationFunction { Point2D(random(-1.0, 1.0), random(0.0, 0.0)) }
 
-        val box = HBox(text, text2)
+        val box = HBox(firstLetter, restOfTheTitle)
         box.alignment = Pos.CENTER
 
         val titleRoot = StackPane(border, box)
-        titleRoot.translateX = appWidth / 2.0 - (textWidth + 30) / 2
+        titleRoot.translateX = appWidth / 2.0 - (titleWidth + 30) / 2
```
AlmasB, line 237: The translateX property should also be bound to the new value of `restOfTheTitle` layout bounds. Something like:

```
firstLetter.layoutBoundsProperty().addListener { _, _, newBounds -> update titleRoot translateX }
```

AlmasB (changes requested):
Thanks for this. It looks good, just need to address the last few bits. Let me know if the attached comments make sense

PR #1400, opened 2025-01-20
feat: Allow sending spawnData when spawning entities using the MultiplayerService

NotBjoggisAtAll:
An attempt at resolving my own question on the [discussion page.](https://github.com/AlmasB/FXGL/discussions/1399#discussion-7850365)

Tested with a local project where I first encountered the need for this.

### DIFF
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/multiplayer/MultiplayerService.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/multiplayer/MultiplayerService.kt
@@ -104,6 +104,10 @@ class MultiplayerService : EngineService() {
     }
 
     fun spawn(connection: Connection<Bundle>, entity: Entity, entityName: String) {
+        spawn(connection, entity, entityName, SpawnData(entity.x, entity.y, entity.z))
+    }
+
+    fun spawn(connection: Connection<Bundle>, entity: Entity, entityName: String, spawnData: SpawnData) {
         if (!entity.hasComponent(NetworkComponent::class.java)) {
             log.warning("Attempted to network-spawn entity $entityName, but it does not have NetworkComponent")
             return
@@ -111,7 +115,7 @@ class MultiplayerService : EngineService() {
 
         val networkComponent = entity.getComponent(NetworkComponent::class.java)
 
-        val event = EntitySpawnEvent(networkComponent.id, entityName, entity.x, entity.y, entity.z)
+        val event = EntitySpawnEvent(networkComponent.id, entityName, NetworkSpawnData(spawnData))
 
         // TODO: if not available
         val data = replicatedEntitiesMap[connection]!!
@@ -129,7 +133,12 @@ class MultiplayerService : EngineService() {
                         val id = event.networkID
                         val entityName = event.entityName
 
-                        val e = gameWorld.spawn(entityName, SpawnData(event.x, event.y, event.z))
+                        val networkSpawnData = event.networkSpawnData
+                        val data = networkSpawnData.bundle.data
+                        val spawnData = SpawnData(networkSpawnData.x, networkSpawnData.y, networkSpawnData.z)
+                        data.forEach(spawnData::put)
+
+                        val e = gameWorld.spawn(entityName, spawnData)
 
                         // TODO: show warning if not present
                         e.getComponentOptional(NetworkComponent::class.java)
--- a/fxgl/src/main/kotlin/com/almasb/fxgl/multiplayer/NetworkSpawnData.kt
+++ b/fxgl/src/main/kotlin/com/almasb/fxgl/multiplayer/NetworkSpawnData.kt
@@ -1,0 +1,36 @@
+/*
+ * FXGL - JavaFX Game Library. The MIT License (MIT).
+ * Copyright (c) AlmasB (almaslvl@gmail.com).
+ * See LICENSE for details.
+ */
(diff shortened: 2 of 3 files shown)


AlmasB:
Thanks for this. The public API seems fine. However, implementation-wise, SpawnData can carry non-serializable data, so I'm just a bit cautious it may break compatibility with existing projects and/or cause confusion regarding the contractual obligation of SpawnData.

Is there a way we can achieve the same without making SpawnData serializable? Perhaps some internal or public data structure (which is serializable) that can carry the spawn data, `NetworkSpawnData extends Serializable`, which contractually forces developers to keep only Serializable data in it, possibly a serialization friendly version of `PropertyMap`.

Another aspect we need to try and figure out is what to do when SpawnData does carry non-serializable data. Using `NetworkSpawnData` from above may solve this. I'm curious to know what your thoughts are on this.

NotBjoggisAtAll:
I agree with your concerns about changing SpawnData serializable, 
I can make a `NetworkSpawnData ` that is serializable instead. I think it should be public so that others can make subclasses to provide more data to send from the server to the client.

I think it should be up the the users to make sure all the data is serializable if they choose to create a subclass of `NetworkSpawnData`

Does this sound okay? I can make the changes and then we can see if its enough

NotBjoggisAtAll:
Where do you want the NetworkSpawnData class to live? Under entity like with SpawnData or together with the MultiplayerService under the multiplayer package?

AlmasB:
Probably alongside MultiplayerService, we can change later as needed. Above all sounds good.

AlmasB:
Not sure if helpful, but perhaps you can make use of `Bundle` to keep the data inside `NetworkSpawnData`. Bundle is fully serializable and is what's used when saving to the file system

NotBjoggisAtAll:
I wonder if `NetworkSpawnData `should extend `SpawnData` since the `GameWorld.spawn()` method expects a object of type `SpawnData`. Otherwise I would need to find a way to convert between the two SpawnData classes

AlmasB:
I'd go with a constructor for now `NetworkSpawnData(SpawnData)`

NotBjoggisAtAll:
I tried to do something like this but my demo fails since SpawnData is not serializable. 

```java
open class NetworkSpawnData(val spawnData: SpawnData) : Serializable {
}
-----------------

    fun spawn(connection: Connection<Bundle>, entity: Entity, entityName: String, spawnData: SpawnData) {
        if (!entity.hasComponent(NetworkComponent::class.java)) {
            log.warning("Attempted to network-spawn entity $entityName, but it does not have NetworkComponent")
            return
        }

        val networkComponent = entity.getComponent(NetworkComponent::class.java)

        val event = EntitySpawnEvent(networkComponent.id, entityName, NetworkSpawnData(spawnData))

        // TODO: if not available
        val data = replicatedEntitiesMap[connection]!!
        data.entities += entity

        fire(connection, event)
    }

----------------------
Demo

     // This does not work.
      SpawnDataspawnData = new SpawnData(x, y);
      Entity entity = spawn("test", spawnData);
      getMultiplayerService().spawn(connection, entity, "test", spawnData);

      //This work because I send my custom spawn data class which implements Serializable
      CardSpawnData cardSpawnData = new CardSpawnData(x, y, card.suit(), card.rank());
      Entity entity = spawn("card", cardSpawnData);
      getMultiplayerService().spawn(connection, entity, "card", cardSpawnData);
   
      
```

I could try to use a Bundle somehow, but I'm not too sure on how to use it. I could also set data inside SpawnData x,y,z,data manually on the NetworkSpawnData, but then I'm not sure to get custom variables/fields that exists on subclasses of SpawnData

AlmasB:
`NetworkSpawnData(val spawnData: SpawnData)` -- the `val` there makes spawnData part of the class. Perhaps get rid of `val` and parse / convert all you need from spawnData's data object into NetworkSpawnData

AlmasB:
I was thinking something like this.

```
NetworkSpawnData(spawnData: SpawnData) {

    val bundle = Bundle("NetworkSpawnData")

    init {
        spawnData.data.forEach((key, value) -> bundle.put(key, value));
    }
}
```

Not sure what we should do if `value` is not serializable. I guess for now you could check if it is, then put(), otherwise ignore.

Once we get NetworkSpawnData through the network to the other end, we need to reconstruct local `SpawnData`, which we should be able to do via the bundle object and its `data` field. Let me know if this makes sense.

Incidentally, I don't know/remember why `SpawnData` is an open class -- its internal Map should be able to carry anything, meaning we can potentially remove the `open` modifier.

NotBjoggisAtAll:
I implemented something like you said, but I also included the position x,y,z because it didn't look like they were a part of `SpawnData::data`. 

I left the `SpawnData` open, you can decide to remove the open modifier at a later stage. I do not know what will break or not if I remove it in this feature.

Does the new changes look like something you had in mind?

AlmasB:
Many thanks, yes that's exactly what I had in mind. I'll take a look and merge later today.

I'll have a think about how to write a simple test for it, but in the meantime, can you confirm this implementation works for your use case?

NotBjoggisAtAll:
Yes, I had to do some adjustments since I extended `SpawnData` expecting that to work. But I was able to retrieve the data in my spawn method from the `spawnData.data` variable instead of my custom class

AlmasB (approved):
Looks good, thank you

PR #1405, opened 2025-02-23
fix: add KeyCode.SHORTCUT to CTRL Input Modifier

### LINKED ISSUE #1204: InputModifier should include SHORTCUT(Ctrl on Win, Command on Mac) modifier?
### Discussed in https://github.com/AlmasB/FXGL/discussions/1203

<div type='discussions-op-text'>

<sup>Originally posted by **chengenzhao** August 27, 2022</sup>
HI:

We just read the source code of FXGL
and in the Triggers.kt
there are some modifiers for the input modifier
and when statement will match Keycode of modifiers to the input modifier
but it only matches KeyCode.Ctrl to InputModifier.Ctrl
I think it would be nice to have ShortCut in that clause also
like:

```kotlin
//KeyTrigger
        return when (event.code) {
            KeyCode.CONTROL, KeyCode.SHORTCUT -> modifier == InputModifier.CTRL
            KeyCode.SHIFT -> modifier == InputModifier.SHIFT
            KeyCode.ALT -> modifier == InputModifier.ALT
            else -> event.code == key
        }

//MouseTrigger
    override fun isReleased(event: InputEvent): Boolean {
        if (event is KeyEvent) {
            return when (event.code) {
                KeyCode.CONTROL, KeyCode.SHORTCUT -> modifier == InputModifier.CTRL
                KeyCode.SHIFT -> modifier == InputModifier.SHIFT
                KeyCode.ALT -> modifier == InputModifier.ALT
                else -> false
            }
        }
```
Because when people use Mac, they tend to use Meta(Command) key rather than Ctrl
so it would be nice to have Short Cut in that clause also

Thanks</div>

abbiereid:
resolves #1204

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/input/Triggers.kt
@@ -98,7 +98,7 @@ data class KeyTrigger
             return false
 
         return when (event.code) {
-            KeyCode.CONTROL -> modifier == InputModifier.CTRL
+            KeyCode.CONTROL, KeyCode.SHORTCUT -> modifier == InputModifier.CTRL
             KeyCode.SHIFT -> modifier == InputModifier.SHIFT
             KeyCode.ALT -> modifier == InputModifier.ALT
             else -> event.code == key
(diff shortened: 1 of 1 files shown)


AlmasB (approved):
Looks good, thanks

PR #1431, opened 2025-09-10
German translation improved

armin-reichert:
Hello Almas,

I saw some strange German translations in this game (https://github.com/kryadov/labyrinth7) and found  that they come from the FXGL framework itself. So I fixed them in the German language file.

Best regards
Armin


AlmasB:
Many thanks @armin-reichert

PR #1455, opened 2025-12-17
build: Bump Kotlin 2.0 -> 2.3, JDK to 23 -> 25, Javafx to 21.0.1 -> 25.0.1, Jackson 2.18.2 -> 2.19.4, JaCoCo 0.8.12 -> 0.8.14, attach 4.0.17 -> 4.0.24

### LINKED ISSUE #1451: Fix vulnerability issues
Image

```
Dependency maven:com.fasterxml.jackson.core:jackson-core:2.14.2 is vulnerable
Dependency maven:org.openjfx:javafx-graphics:21.0.1 is vulnerable
Dependency maven:org.openjfx:javafx-media:21.0.1 is vulnerable
```

chengenzhao:
Bump Kotlin to 2.3, JDK to 25, JavaFX to 25.0.1

Upgrade Kotlin compiler jvm target to 21 instead of 17(onle one LTS version lower than current LTS version) 

Modify deprecated kotlin API String.toLowercase() to .lowercase() which is required modification for upgrading Kotlin version

this PR should fix #1451


chengenzhao:
@AlmasB there is one test failed

```Kotlin
    fun `Path`() {
        val rect = Rectangle(160.0, 50.0)

        val anim = AnimatedPath(rect)

        assertThat(anim.getValue(0.0), `is`(Point3D(0.0, 0.13125, 0.0)))
        assertThat(anim.getValue(1.0), `is`(Point3D(0.0, 0.0, 0.0)))
        assertThat(anim.getValue(0.5), `is`(Point3D(160.0, 50.0, 0.0)))
    }
```

java.lang.AssertionError: 

Expected: is <Point3D [x = 0.0, y = 0.13125, z = 0.0]>
     but: was <Point3D [x = 0.0, y = 0.0, z = 0.0]>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at com.almasb.fxgl.animation.AnimatedValueTest.Path(AnimatedValueTest.kt:110)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)

prompting 

> WARNING: A restricted method in java.lang.System has been called
> WARNING: java.lang.System::load has been called by com.sun.glass.utils.NativeLibLoader in module javafx.graphics (file:/Users/chengenzhao/.m2/repository/org/openjfx/javafx-graphics/25.0.1/javafx-graphics-25.0.1-mac-aarch64.jar)
> WARNING: Use --enable-native-access=javafx.graphics to avoid a warning for callers in this module
> WARNING: Restricted methods will be blocked in a future release unless native access is enabled

seems like it is the old JNI method should use new FFM api to replace it?

chengenzhao:
@AlmasB there is one test failed

```Kotlin
    fun `Path`() {
        val rect = Rectangle(160.0, 50.0)

        val anim = AnimatedPath(rect)

        assertThat(anim.getValue(0.0), `is`(Point3D(0.0, 0.13125, 0.0)))
        assertThat(anim.getValue(1.0), `is`(Point3D(0.0, 0.0, 0.0)))
        assertThat(anim.getValue(0.5), `is`(Point3D(160.0, 50.0, 0.0)))
    }
```

Got the solution here
```kotlin
        while (t < 1.0) {
            points[percent++] = Point3D(dummy.translateX, dummy.translateY, 0.0)

            t += 0.01
            Thread.sleep(10)//add this line, seems like if the time elapsed speed is not matching the transition time, the calculation wont activated...

            pt.jumpTo(Duration.seconds(t))
        }
```

chengenzhao:
Here is the recreation of the bug:
```java

  public static void main(String... args) throws Exception {

    var c = new Circle(10,Color.RED);
    var pt = new PathTransition(Duration.seconds(1), new Rectangle(160,50), c);
    pt.play();

    Thread.sleep(500);//comment out this line, translate X wont changed
    pt.jumpTo(Duration.seconds(0.5));

    IO.println(c.getTranslateX());//print 160.0, but if comment out the above line, here will be 0.0
  }

```

dont know exactly the reason but it works

fxgl-controllerinput/src/main/kotlin/com/almasb/fxgl/controllerinput/GameController.java
AlmasB, line 1: This file shouldn't be part of PR

fxgl-core/src/main/kotlin/com/almasb/fxgl/animation/AnimatedValue.kt
AlmasB, line 1: I do not recall previous tests failing, is this a result of the new PR?

fxgl-core/src/main/kotlin/com/almasb/fxgl/core/Copyable.java
AlmasB, line 1: This file shouldn't be part of PR

AlmasB (commented):
Thanks for the version bumps, there seem to be a few unrelated changes. Can you have a look at the comments please.

fxgl-core/src/main/kotlin/com/almasb/fxgl/animation/AnimatedValue.kt
chengenzhao: I think it is because when we upgrade JavaFX version to 25 then it causes this problem
I have recreate the bug by the Java code and test it in JavaFX 25.0.1
and when it failed, we directly run the test and it failed also then here is the recreation of the problem

```java
  public static void main(String... args) throws Exception {

    var c = new Circle(10,Color.RED);
    var pt = new PathTransition(Duration.seconds(1), new Rectangle(160,50), c);
    pt.play();

    Thread.sleep(500);//comment out this line, translate X wont changed
    pt.jumpTo(Duration.seconds(0.5));

    IO.println(c.getTranslateX());//print 160.0, but if comment out the above line, here will be 0.0
  }
```

basically this code should work as expected 
```java
var pt = new PathTransition(Duration.seconds(1), new Rectangle(160,50), c);
pt.play();
pt.jumpTo(Duration.seconds(0.5));
assertEqual(c.getTranslateX() == 160.0);
```

but it failed
it is 0.0, the circle node didnt move at all
but if we add in like this line 
```java
Thread.sleep(500);
```
then it works as epected

and I noticed that the CI failed in all three OS(win,mac&linux)
so you may recreate this problem in your machine just using JavaFX 25.0.1

chengenzhao: I just tested the same code using JavaFX 21.0.1
it works as expected no need to use Thread.sleep
so pretty sure it is the problem of JavaFX 
when we upgrade JavaFX from 21.0.1 to 25.0.1 
it happens

so any suggestions?

chengenzhao: As you may see here, using JavaFX 21.0.1 it works fine
image
this is JavaFX 25.0.1 assert failed
image

README.md
```diff
@@ -79,7 +79,7 @@ Otherwise, see:
 <dependency>
     <groupId>com.github.almasb</groupId>
     <artifactId>fxgl</artifactId>
-    <version>21.1</version>
+    <version>25</version>
```
AlmasB: FXGL 25 is not released yet, so should be 21.1. This will change to 25 automatically as part of the release process.

fxgl-core/src/main/kotlin/com/almasb/fxgl/animation/AnimatedValue.kt
```diff
@@ -163,6 +163,7 @@ class AnimatedPath
             points[percent++] = Point3D(dummy.translateX, dummy.translateY, 0.0)
 
             t += 0.01
+            Thread.sleep(10)//seems like if the time elapsed speed is not matching the transition time, the calculation wont be activated
```
AlmasB: Ok, thanks for investigating. This is rarely used, so it sounds fine to have this workaround. Please can you add a comment that this change is due to JavaFX behaviour being different between versions 21 and 25.

README.md
chengenzhao: ok

AlmasB (approved):

PR #1456, opened 2026-01-09
fix: bind fullscreen button text to localization service

Shimijp:
The fullscreen button text was set only once during initialization.
Changed it to bind to the localization property so it updates dynamically when language changes.
(Followed the suggested fix)


AlmasB (approved):
Many thanks for this

PR #1459, opened 2026-01-23
adding getDistance helper to FXGLMath.java

alexpipilas:
Hello guys ! I have added a new helper method `getDistance` to the `FXGLMath` class. This method provides a standard way to calculate the distance between two 2D points, which is a common requirement in game development.

- Implemented `getDistance(double x1, double y1, double x2, double y2)`.
- Added comprehensive Javadoc to explain parameters and return values.

Centralizing this calculation in `FXGLMath` promotes code reuse and makes the engine's math utilities more complete for developers building 2D games.


AlmasB:
Thank you for this, but there is already a standard way to represent points in 2D and 3D spaces in JavaFX with existing distance functions:

```
var p1 = new Point2D(x1, y1);
var p2 = new Point2D(x2, y2);

p1.distance(p2);
```

PR #1465, opened 2026-02-12
fix: renamed json with conflicting name

tomocrea:
Renamed simple.json in fxgl-samples to simplesample.json to fix test Load DialogueGraph that was meant to search for another json with the same name


AlmasB (approved):
Looks good, thank you

Issue #1439, opened 2025-09-26
Timer Component


AlmasB:
### Discussed in https://github.com/AlmasB/FXGL/discussions/1434

<div type='discussions-op-text'>

<sup>Originally posted by **LifeDrainingCoding** September 17, 2025</sup>
I wonder , why FXGL doesn't have Timer Component that will schedule task  by given interval? 

This will allows devs to execute tasks to each entity individually at individually calculated interval.

My use case: I have a `Marble` that's shoots projectiles , and when this projectile collides with another `Marble` that will lower it's radius or disappear. Shoot rate of `Marble` depends on it's radius, so i need update interval every radius change.

```Java
import com.almasb.fxgl.dsl.FXGL;
import com.almasb.fxgl.entity.Entity;
import com.almasb.fxgl.entity.SpawnData;
import com.almasb.fxgl.entity.components.BoundingBoxComponent;
import com.almasb.fxgl.entity.components.CollidableComponent;
import com.almasb.fxgl.entity.components.ViewComponent;
import com.almasb.fxgl.physics.BoundingShape;
import com.almasb.fxgl.physics.HitBox;
import com.almasb.fxgl.physics.PhysicsComponent;
import com.almasb.fxgl.physics.box2d.dynamics.BodyDef;
import com.almasb.fxgl.physics.box2d.dynamics.BodyType;
import com.almasb.fxgl.physics.box2d.dynamics.FixtureDef;
import com.lifedrained.fxglphysicstest.FXGLTest;
import com.lifedrained.fxglphysicstest.components.TimerComponent;
import com.lifedrained.fxglphysicstest.types.TypeOf;
import javafx.scene.paint.Color;
import javafx.scene.shape.Circle;
import javafx.util.Duration;

import java.util.UUID;

import static com.lifedrained.fxglphysicstest.SpawnDataKeys.*;

public class Marble extends Entity {

    protected boolean isShooting =false;
    protected PhysicsComponent physics;
    protected BodyDef bodyDef;
    protected FixtureDef fixtureDef;
    protected ViewComponent view;
    protected BoundingBoxComponent bbox;
    protected TimerComponent timer = new TimerComponent();
    protected double radius, mass,area , density = 0;

    public Marble(SpawnData data) {
        super();
        setX(data.getX());
        setY(data.getY());
        setType(TypeOf.MARBLE_TYPE);
        radius = data.get(RADIUS);
        mass = data.get(MASS);

        bodyDef = new BodyDef();
[...]

PR #1469, opened 2026-03-23
feat: TimerActionComponent to schedule actions per entity.

michqql:
Addresses issue #1439.

### Summary
`Component` that wraps a `Timer` object and updates timer actions dependent on the entities time per frame value. This component can be used to schedule actions with either a delay or interval, with overloaded functions matching the underlying `Timer` class.

A `PongVsComputerApp` sample is included to see example usage of the `TimerActionComponent`.

### Features
Includes the same overloaded methods as `Timer`:
- `runAtInterval(Runnable action, Duration interval)`
- `runAtInterval(Runnable action, Duration interval, int limit)`
- `runAtIntervalWhile(Runnable action, Duration interval, ReadOnlyBooleanProperty whileCondition)`
- `runOnceAfter(Runnable action, Duration delay)`
- `clear()`

### Testing
`TimerActionComponentTest` includes 12 tests to verify the actions are executed. However these tests try not to duplicate `TimerActionTest`.

### Example Usage
(From the sample)
```
TimerActionComponent timerComponent = new TimerActionComponent();

/* Move the Y position of the computer bat to the Y position of the ball */
timerComponent.runAtInterval(() -> {
    Entity ball = timerComponent.getEntity().getObject("ballEntity");
    timerComponent.getEntity().setY(ball.getY() - timerComponent.getEntity().getHeight() / 2f);
}, Duration.seconds(1 / 60f));

entityBuilder()
        .at(x, y)
        .viewWithBBox(new Rectangle(PADDLE_WIDTH, PADDLE_HEIGHT))
        .with(timerComponent)
        .with("ballEntity", ball)
        .buildAndAttach();
```

### DIFF
--- a/fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
+++ b/fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
@@ -13,18 +13,19 @@ import javafx.util.Duration;
 
 /**
  * Component to schedule the execution of actions.
+ * The timer's time per frame (TPF) is tied to the entity's TPF.
+ * The entity's TPF can be modified by {@link com.almasb.fxgl.entity.components.TimeComponent}.
  *
  * @implNote - A wrapper around a Timer class, as a component.
  *
  * @author Michael Pearson (<a href="https://github.com/michqql/">https://github.com/michqql/</a>)
  */
-public class TimerActionComponent extends Component {
+public final class TimerActionComponent extends Component {
 
     private final Timer timer = new Timer();
 
     @Override
     public void onUpdate(double tpf) {
-        super.onUpdate(tpf);
         this.timer.update(tpf);
     }
 
--- a/fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
+++ b/fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
@@ -1,0 +1,86 @@
+/*
+ * FXGL - JavaFX Game Library. The MIT License (MIT).
+ * Copyright (c) AlmasB (almaslvl@gmail.com).
+ * See LICENSE for details.
+ */
+
+package com.almasb.fxgl.entity.component;
+
+import com.almasb.fxgl.time.Timer;
+import com.almasb.fxgl.time.TimerAction;
+import javafx.beans.property.ReadOnlyBooleanProperty;
+import javafx.util.Duration;
+
+/**
+ * Component to schedule the execution of actions.
+ *
+ * @implNote - A wrapper around a Timer class, as a component.
+ *
+ * @author Michael Pearson (<a href="https://github.com/michqql/">https://github.com/michqql/</a>)
+ */
+public class TimerActionComponent extends Component {
+
+    private final Timer timer = new Timer();
+
+    @Override
+    public void onUpdate(double tpf) {
+        super.onUpdate(tpf);
+        this.timer.update(tpf);
+    }
+
+    /**
+     * The Runnable [action] will be scheduled to start at given [interval].
+     * The action will start for the first time after given interval.
+     * The action will be scheduled unlimited number of times unless user cancels it
+     * via the returned action object.
+     *
+     * @return timer action
+     */
+    public TimerAction runAtInterval(Runnable action, Duration interval) {
+        return timer.runAtInterval(action, interval);
+    }
+
+    /**
+     * The Runnable [action] will be scheduled to start at given [interval].
+     * The action will start for the first time after given interval.
+     * The action will be scheduled [limit] number of times unless user cancels it
+     * via the returned action object.
+     *
+     * @return timer action
+     */
+    public TimerAction runAtInterval(Runnable action, Duration interval, int limit) {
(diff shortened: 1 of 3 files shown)


fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
```diff
@@ -0,0 +1,86 @@
...
+import javafx.util.Duration;
+
+/**
+ * Component to schedule the execution of actions.
+ *
+ * @implNote - A wrapper around a Timer class, as a component.
+ *
+ * @author Michael Pearson (<a href="https://github.com/michqql/">https://github.com/michqql/</a>)
+ */
+public class TimerActionComponent extends Component {
```
AlmasB: Should be fine to make it `final` by default. It can be opened in the future if required

fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private final Timer timer = new Timer();
+
+    @Override
+    public void onUpdate(double tpf) {
+        super.onUpdate(tpf);
```
AlmasB: `super.onUpdate(tpf);` calls Component onUpdate(), which I think may be empty. If it is, then we can safely remove this line.

fxgl-entity/src/main/java/com/almasb/fxgl/entity/component/TimerActionComponent.java
```diff
@@ -0,0 +1,86 @@
...
+
+package com.almasb.fxgl.entity.component;
+
+import com.almasb.fxgl.time.Timer;
+import com.almasb.fxgl.time.TimerAction;
+import javafx.beans.property.ReadOnlyBooleanProperty;
+import javafx.util.Duration;
+
+/**
+ * Component to schedule the execution of actions.
```
AlmasB, line 15: Worth noting that the timer's clock will be tied to the entity's clock. So 1 second in this TimerActionComponent will depend on how long 1 second is for the entity, e.g. TimeComponent may alter that behaviour (which is the correct desirable behaviour). Not quite sure exactly the wording for the doc, but something along these lines should do the trick.

fxgl-entity/src/test/java/com/almasb/fxgl/entity/component/TimerActionComponentTest.java
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * @author Michael Pearson (<a href="https://github.com/michqql/">https://github.com/michqql/</a>)
+ */
+public class TimerActionComponentTest {
+
+    /* Counter class to allow for atomic operations from Runnable action */
+    private static class Counter {
+        private int value;
+
+        int  get      () { return value    ; }
```
AlmasB: This formatting is largely inconsistent with the rest of the project's formatting. This will need to be updated to follow the standard Java guidelines. Though I think the fix for the above comment will remove the need for this altogether.

fxgl-entity/src/test/java/com/almasb/fxgl/entity/component/TimerActionComponentTest.java
```diff
@@ -0,0 +1,240 @@
...
+        void reset    () {        value = 0; }
+        void increment() {        value++  ; }
+    }
+
+    private final Counter  executionCounter = new Counter();
+    private final Runnable action           = executionCounter::increment;
```
AlmasB: Same comment about formatting Line 34 and 35 as above. There is no need to align elements.

fxgl-entity/src/test/java/com/almasb/fxgl/entity/component/TimerActionComponentTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Tests for the {@link TimerActionComponent} class.
+ *
+ * @author Michael Pearson (<a href="https://github.com/michqql/">https://github.com/michqql/</a>)
+ */
+public class TimerActionComponentTest {
+
+    /* Counter class to allow for atomic operations from Runnable action */
+    private static class Counter {
```
AlmasB: It's fine to use such a helper class, but best to avoid using the same object from multiple tests in this way. For example, if the test environment runs the tests in parallel (I don't think it does, however we should try best practice where it's easy to do so), there may be an issue with multiple threads writing and reading the same data.

This would mean simply using an `int count = 0`. If this doesn't work because of lambda functions, you can use `var count = new SimpleIntegerProperty()`, in each test (like you have done in `public void testRunAtIntervalConditionalCancelledEarly()`). This should give you more or less the same functionality, but in a relatively safer way.

AlmasB (commented):
Many thanks for this PR, it looks good. 

1) Please can you check the comments below.
2) Also the checks seem to be failing, could you take a look please.

michqql:
Commit today:
- Made TimerActionComponent class final
- Removed call to super#onUpdate in TimerActionComponent
- Reworded documentation in TimerActionComponent class
- Fixed formatting in TimerActionComponentTest
- Removed member variables from TimerActionComponentTest to support parallelism

PR #1470, opened 2026-03-31
fix: recycle particle instead of emitter position back to pool in ParticleSystem

### LINKED ISSUE #1417: BUG: In the update method of the ParticleSystem, the particle objects are not correctly recycled back to the object pool. Here, the incorrect reference p of Point2D is used for the recycling process.
> com.almasb.fxgl.particle.ParticleSystem

```
override fun onUpdate(tpf: Double) {
        emitters.forEach { (emitter, p) ->
            val particlesList = particles[emitter]!!

            particlesList.addAll(emitter.emit(p.x, p.y))

            val iter = particlesList.iterator()
            while (iter.hasNext()) {
                val particle = iter.next()

                if (particle.update(tpf)) {
                    iter.remove()

                    pane.children.remove(particle.view)
                    Pools.free(p)  //# Here, the incorrect objects have been retrieved.
                } else {
                    if (particle.view.parent == null)
                        pane.children.add(particle.view)
                }
            }
        }
    }
```

evanofficial:
Fixes #1417

In `ParticleSystem.onUpdate`, when a particle expires it should be returned to the object pool via `Pools.free(particle)`. Instead, `Pools.free(p)` was being called, which freed the emitter's `Point2D` position — a completely different object.

Changed `Pools.free(p)` → `Pools.free(particle)` on the removal path.

### DIFF
--- a/fxgl-core/src/main/kotlin/com/almasb/fxgl/particle/ParticleSystem.kt
+++ b/fxgl-core/src/main/kotlin/com/almasb/fxgl/particle/ParticleSystem.kt
@@ -48,7 +48,7 @@ class ParticleSystem : Updatable {
                     iter.remove()
 
                     pane.children.remove(particle.view)
-                    Pools.free(p)
+                    Pools.free(particle)
                 } else {
                     if (particle.view.parent == null)
                         pane.children.add(particle.view)


AlmasB (approved):
Thanks!

Issue #1433, opened 2025-09-16
Expose MassData to reduce boilerplate code


AlmasB:
### Discussed in https://github.com/AlmasB/FXGL/discussions/1432

<div type='discussions-op-text'>

<sup>Originally posted by **LifeDrainingCoding** September 15, 2025</sup>
Is there in FXGL 21 any way to configure `Entity` mass except adding fixture definitions with density? Btw, `MassData` is located in package that's not visible. I think it's essential to expect mass property of physical bodies in game engine but i didn't find any except density. </div>

Issykelly:
Two new methods could be added to `PhysicsComponent`: `setMassData` and `getMassData`. These would allow more direct control by the developer when dealing with entities and their physics. 

Currently, in FXGL, there is no way to configure the Entity mass directly. Instead, mass is derived from fixture density, which determines the result of the underlying mass calculations. This follows the `Box2D` design, but it can make gameplay tuning less intuitive because it allows less direct control for the developer. This would be especially frustrating when working with multiple entities or when a specific mass value is desired.

Since `MassData` already exists in the underlying physics implementation, exposing it could provide a more direct way to configure mass and reduce the need for manual density adjustments.

By adding a `setMassData` method to `PhysicsComponent`, which would require importing `com.almasb.fxgl.physics.box2d.collision.shapes.MassData`, we could allow developers more direct control. Additionally, adding `getMassData` could help speed up debugging.

AlmasB:
Thanks @Issykelly 
Given `MassData` already exists as part of the implementation, we could expose the API publicly, rather than adding extra methods. I don't remember where `MassData` is located. It's either in Body or Fixture. I think the direct solution would be:

- make MassData public if it isn't / make module-info export the appropriate package
- make the set and get accessors public
- add a test if there isn't one

Does the above make sense?

Issykelly:
Thank you for the quick response! That makes sense, I've found `MassData` and added it to the module-info exports - it was already public. 

There is also already a public `getMass` method within Body, but no `setMass`. Instead, there's a `getMassData` method which returns the entire mass object (which includes the mass, centre and the rotational inertia) and there's a `setMassData`. These are both already public, and whilst this does allow users to directly interact with the mass data, it requires them to input a `MassData` object, which requires more than just the mass value and may not be as easy to implement as developers would like.

[...]

PR #1472, opened 2026-04-10
added new sample code to demonstrate mass

Issykelly:
#1433 I've added a new sample to demonstrate how mass affects physics, as well as how to directly make changes to the mass of an object. This demo uses small balls and seesaws to showcase how objects of different masses interact with each other and their environment. 

Use the "Q", "E", "O" & "P" buttons to affect the on-screen mass. Then, click the left mouse button to spawn balls with the shown mass. Use these balls to interact with the seesaws and other balls and see how they interact!

As well as this, I've added the `MassData` code to the module info exports.

### DIFF
--- a/fxgl-samples/src/main/java/basics/MassSample.java
+++ b/fxgl-samples/src/main/java/basics/MassSample.java
@@ -7,7 +7,6 @@
 
 package basics;
 
-
 import com.almasb.fxgl.app.GameApplication;
 import com.almasb.fxgl.app.GameSettings;
 import com.almasb.fxgl.dsl.FXGL;
@@ -28,9 +27,7 @@ import javafx.scene.shape.Rectangle;
 import javafx.scene.text.Font;
 import javafx.scene.text.Text;
 
-
 import static com.almasb.fxgl.dsl.FXGL.*;
-
 
 /**
  * Shows how to change mass within a body directly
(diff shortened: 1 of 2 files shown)


Issykelly:
#1433

fxgl-samples/src/main/java/basics/MassSample.java
```diff
@@ -0,0 +1,213 @@
...
+import com.almasb.fxgl.physics.box2d.dynamics.FixtureDef;
+import javafx.geometry.Point2D;
+import javafx.scene.input.KeyCode;
+import javafx.scene.input.MouseButton;
+import javafx.scene.paint.Color;
+import javafx.scene.shape.Circle;
+import javafx.scene.shape.Rectangle;
+import javafx.scene.text.Font;
+import javafx.scene.text.Text;
+
```
AlmasB, line 29: It seems your IDE is adding an extra new line for each empty line, please can you check. Standard practice is 1 empty line, like at Line 38.

AlmasB (commented):
Thanks for this, it looks good. Just one comment below.

fxgl-samples/src/main/java/basics/MassSample.java
Issykelly: Thank you for the response! I've gone ahead and fixed this. I'll be mindful to ensure I don't make the same formatting error again :)

AlmasB (approved):
Many thanks, looks good.

